PR #14, opened 2024-09-27
Prefect build

andhus:
Adds logic for building a `stardag` DAG using `prefect` primitives:
- wraps each stardag.Task in an prefect.task
- chains dependencies via `wait_for` (to get dependencies displayed in UI)
- PoC of artifacts creation via callbacks.


andhus:
Screenshot

image

PR #110, opened 2026-03-20
Refactor DagGraph: extract graph builders into testable functions

andhus:
## Summary
- Extract `buildGraphNodes()` and `buildGraphEdges()` from the large `useMemo` into standalone functions with typed parameter objects
- Rename `getLayoutedElements` → `applyDagreLayout` for clarity
- Add `getHandlePositions()` helper that derives source/target positions from direction, setting `sourcePosition`/`targetPosition` on nodes
- The `useMemo` now reads as three clear steps: build nodes → build edges → apply layout

Stacked on #109 — merge that first.

## Test plan
- [ ] Verify DAG rendering is visually identical (no layout/styling changes)
- [ ] Verify node click, drag, direction toggle, reset layout, fullscreen all still work
- [ ] Verify edge animations on running tasks still work

### DIFF
--- a/app/stardag-ui/src/components/DagGraph.tsx
+++ b/app/stardag-ui/src/components/DagGraph.tsx
@@ -4,6 +4,7 @@ import {
   Controls,
   type Node,
   type Edge,
+  Position,
   useNodesState,
   useEdgesState,
   type NodeTypes,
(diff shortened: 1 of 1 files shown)


andhus:
[By Claude] Closing this unmerged, six months after it was opened. Nothing is wrong with the refactor — it just is not salvageable as a diff any more:

- It is stacked on `long-names-in-dag-view-fix` (#109), which merged the same day. GitHub still reports it mergeable and clean, but that is against a base branch that no longer means anything — and because the CI and Copilot-review rulesets are both scoped to the default branch, it never got a check run or a review either.
- `DagGraph.tsx` has moved on twice since, in #123 (dynamic dep edges) and #166 (Modal execution surfacing). Against this PR's base the file is +81/-2, and most of that lands in the exact block this PR rewrites.

The intent is not lost: STA-61 carries it forward — extract `buildGraphNodes()`/`buildGraphEdges()` out of the `useMemo`, rename `getLayoutedElements` -> `applyDagreLayout`, add `getHandlePositions()`, and put a vitest suite behind the result (there is still no `DagGraph.test.tsx`). Re-derived against current `main` rather than rebased.

The `dag-view-deep-refactor` branch stays, so this diff remains readable as the reference for that work.

PR #126, opened 2026-04-27
API perf Tier 2: JSONB columns, env+created indices, batched dep reconciliation

andhus:
## Summary

Tier 2 of the API perf rescue — DB schema improvements and the dep-registration hot path. See **#125** for the full diagnosis (production CPU saturation under SDK build bursts on 2026-04-27). Best merged after #125 (Tier 1) but is **independent** of it (no overlapping files).

### Three related changes

1. **JSONB column types** — `tasks.task_data`, `events.event_metadata`, `task_artifacts.body_json`, `builds.root_task_ids` change from `JSON` (text) to `JSONB`. Postgres now parses these once on write and keeps a binary representation; ` ->>` no longer reparses on every row scan. Eliminates a major contributor to slow Task Explorer JSONB queries on production-scale data.

   Models declare `JSON().with_variant(JSONB(), "postgresql")` so SQLite tests keep `JSON` and Postgres uses `JSONB`. The migration uses `ALTER COLUMN ... TYPE jsonb USING ...::jsonb` (one-pass rewrite). Round-tripped against local Postgres 16: upgrade → downgrade → upgrade clean.

2. **New btree indices**
   - `ix_tasks_environment_created (environment_id, created_at)`
   - `ix_task_artifacts_environment_created (environment_id, created_at)`

   Covers the most common Task Explorer / `/tasks/keys` / `/tasks/values` shape: filter by environment, sort by `created_at DESC`. Postgres can scan the btree in either direction so explicit DESC isn't needed.

3. **Batched dependency reconciliation** — `routes/builds.py:_reconcile_dependency_edges` previously did N round-trips per upstream id (SELECT existing → INSERT phantom → SELECT phantom → INSERT edge). Now does **exactly 3 statements** regardless of N:

   1. SELECT existing tasks WHERE `task_id IN (...)`.
   2. Bulk INSERT phantoms with ON CONFLICT DO NOTHING + one re-fetch.
   3. Bulk INSERT edges with ON CONFLICT DO NOTHING.

   Still idempotent under concurrent registrations; first-write-wins on `is_dynamic` is preserved. A 5-dep `register_task` drops from ~15 RT to 3 RT — meaningful on the SDK build-burst hot path.

### Test infrastructure

The pre-existing `pg_engine` fixture had a latent bug (alembic env.py uses `asyncio.run` which deadlocks inside an async pytest fixture's event loop). Fixed by switching to `Base.metadata.create_all`, matching the SQLite fixture's behaviour. Migrations themselves are exercised via the docker-compose pipeline. Added `pg_session` helper fixture.

### New tests

`tests/test_dependency_reconciliation.py` — 5 Postgres-only integration tests covering: missing-only / existing-only / mixed / empty / duplicate inputs, plus idempotency under repeated calls. Run with `STARDAG_API_TEST_DATABASE_URL=postgresql+asyncpg://... pytest`.

### Verification

- `pytest` on SQLite: 191 passed, 6 skipped (5 new PG-only + 1 prior).
- `pytest` with `STARDAG_API_TEST_DATABASE_URL` set to a local Postgres 16: 196 passed, 1 skipped.
- Migration upgrade and downgrade against the same local DB: clean.
- `pre-commit run --files <changed>`: clean (ruff, pyright, prettier).

### Production rollout

Merging this triggers a deploy. The migration's biggest cost is the JSONB rewrite; observed prod sizes are tiny (events 1672 kB heap, tasks 2272 kB), so it's seconds. The migration runs in transactional DDL (Postgres default for alembic) so a failure rolls back cleanly.

## Test plan

- [ ] CI: stardag-api SQLite pytest passes
- [ ] (If wired up) CI: PG pytest passes
- [ ] Reviewer skim of `_reconcile_dependency_edges` batching (ON CONFLICT semantics, race-loser re-fetch, KeyError on concurrent delete is intentional)
- [ ] Reviewer skim of migration's `_jsonb_variant` alias and that downgrade is the inverse
- [ ] Confirm the model-level `JSON().with_variant(JSONB, "postgresql")` shape is acceptable (vs. switching the underlying type wholesale)


andhus (commented):
## Copilot review responses

All 10 comments addressed in c69a7c8. Per-comment status below; line numbers reference the new commit.

### 1. `routes/builds.py:734` — `len(edge_rows)` fallback over-counts under conflicts

**Resolved.** Changed the fallback to `0` (genuinely conservative — never over-claims). Updated the comment and the function-level docstring's return contract to match. On Postgres + asyncpg the cursor rowcount is reliable, so this fallback path is theoretical, but `AddDependenciesResponse.added` will now never over-state on a dialect that omits rowcount.

### 2 & 3 & 9. Migration: add explicit `postgresql_using=...`

**Resolved.** All four `op.alter_column` calls in `upgrade()` now pass `postgresql_using="<col>::jsonb"`; the four in `downgrade()` pass `<col>::json`. SQLAlchemy's autogenerated DDL had been emitting an implicit cast that worked on Postgres 16 in my local round-trip, but explicit `USING` is version-independent and unambiguous. Round-tripped (upgrade → downgrade → upgrade) on Postgres 16 again after the change.

### 4. `_reconcile_dependency_edges` docstring claimed "three statements" but does four when phantoms are missing

**Resolved.** Rewrote the docstring to state honestly: **two** statements when every upstream task already exists, **four** when missing phantoms must be created (SELECT existing → INSERT phantoms → SELECT re-fetch → INSERT edges). Did *not* take the suggestion to always-INSERT-all-ids — keeping the existing-tasks fast path because most repeat-builds skip the phantom-insert and re-fetch entirely.

### 5. `pg_engine(request)` — unused parameter

**Resolved.** Dropped the unused `request` parameter from the fixture signature.

### 6, 7, 8. Test SELECTs scoped only by `task_id`

**Resolved.** All three (`test_creates_phantoms_and_edges_in_one_call`, `test_mixed_existing_and_missing_upstreams`, `test_duplicate_ids_in_input_are_deduplicated`) now also filter by `Task.environment_id == DEFAULT_ENVIRONMENT_ID`, matching the `(environment_id, task_id)` uniqueness constraint and future-proofing the tests against fixture-data growth.

### 10. `_count_edges_into` redundantly wraps `result.scalars().all()` in `list(...)`

**Resolved.** Simplified to `len(result.scalars().all())`.

---

**Verification:** migration round-tripped on local Postgres 16 with new explicit USING clauses. `pytest` SQLite 191 passed, Postgres 196 passed (including the 5 dep-reconciliation integration tests). `pre-commit` clean.

andhus (commented):
## Review responses

All concerns addressed in commit b690142.

### 1. Migration coverage in `pg_engine` — RESTORED

The earlier `Base.metadata.create_all` workaround is replaced with real alembic migrations:

- `migrations/env.py` now respects `sqlalchemy.url` passed via `alembic.Config` (e.g. by a test fixture) instead of unconditionally overwriting it. Production deploys are unaffected — `alembic.ini` leaves it empty.
- `tests/conftest.py:pg_engine` drops+recreates the `public` schema (the initial migration's autogenerated downgrade doesn't drop `workspacerole` / `invitestatus` enum types), then runs `alembic upgrade head` via `asyncio.to_thread` so env.py's internal `asyncio.run` doesn't collide with the fixture's event loop.
- New **`test-python-pg`** CI job spins a Postgres service and runs `tox` with `STARDAG_API_TEST_DATABASE_URL` set. PG-only tests now fire on every PR.

### 2. Migration locks at scale — DOCUMENTED

Added a `LOCK / DOWNTIME NOTE` to the migration docstring explaining `ALTER COLUMN` takes `ACCESS EXCLUSIVE` and `op.create_index` takes `SHARE`. Sized for current data; would need `CREATE INDEX CONCURRENTLY` + column-swap at 100x growth.

### 3. KeyError comment — TIGHTENED

Now says the lookup is essentially unreachable under READ COMMITTED + ON CONFLICT DO NOTHING; the realistic concurrent-delete failure mode is an FK violation on the edge insert. Lookup stays as a tripwire.

### 4. Postgres-only path on SQLite — handled by #1

New CI job runs the PG-only tests on every PR.

### 5. Concurrency test — ADDED

`test_concurrent_reconcile_overlapping_upstreams_resolves_safely` fires two reconciliations on overlapping missing upstream ids via `asyncio.gather`, asserts exactly 3 phantoms + 3 edges per downstream.

### Minor

- Double `now = utc_now()` collapsed.
- `_existing_json` kept for symmetry with `_jsonb_variant` (your second option).
- `pg_session` fixture left as a general utility.

### Verification

- pytest SQLite: 191 passed, 7 skipped (PG-only).
- pytest with `STARDAG_API_TEST_DATABASE_URL`: 197 passed, 1 skipped.
- Migration round-tripped on local Postgres 16. pre-commit clean.

PR #128, opened 2026-04-28
Fix /tasks/search 500 on build_id filter (UUID cast)

andhus:
## Summary

Surfaced from production logs at 2026-04-28T07:23 UTC during post-Tier-1 sanity-checking:

```
GET /api/v1/tasks/search?...&filter=build_id:=:<uuid>  →  500
asyncpg.exceptions.UndefinedFunctionError: operator does not exist:
uuid = character varying
```

`builds.id` is a Postgres UUID column. The search filter binds the value from the URL query string as text; asyncpg sends it as `varchar`; Postgres rejects the implicit `uuid = varchar` comparison without an explicit cast. Independent of the API perf tiers — same bug pre- and post-deploy.

The same code path was also broken for `ILIKE` (substring match): `uuid ILIKE varchar` is also rejected. The user just hadn't hit it because the production UI uses `=`.

## Fix

`routes/search.py:build_jsonb_condition` for `key == "build_id"`:
- **Equality / inequality** (`=`, `!=`): emit `builds.id <op> CAST(:p AS uuid)` — the bound parameter is coerced to uuid so Postgres can compare against the uuid column. PK index still usable.
- **Substring** (`ILIKE`): emit `builds.id::text ILIKE :p` — the column is rendered as text first.

## Tests

- **`TestBuildIdFilterSQL`** (3 unit tests): pin the emitted SQL strings. Run on every backend (no DB needed). Tripwires for silent regressions.
- **`test_search_filter_by_build_id_equality_does_not_500`** — end-to-end via a new `pg_client` fixture (added to `conftest.py` — the Postgres-equivalent of the existing `client` fixture). Creates a build + task and asserts `filter=build_id:=:<uuid>` returns 200.
- **`test_search_filter_by_build_id_ilike_works_against_uuid_column`** — same shape but with `~` (ILIKE), confirming the column-to-text cast.

Verified the tests would catch the regression by temporarily reverting the `search.py` change and re-running: **all 5 tests fail without the cast, all 5 pass with it**.

The search route is otherwise Postgres-specific (raw JSONB operators, etc.) so the e2e tests skip on SQLite via the `pg_client` fixture's underlying `pg_engine` skip-when-unset behaviour. The new unit tests run on every backend.

## Verification

- pytest SQLite: 210 passed, 9 skipped (all PG-only).
- pytest with `STARDAG_API_TEST_DATABASE_URL`: 218 passed, 1 skipped.
- pre-commit clean (ruff, pyright, prettier).
- The new `Test Python (stardag-api on Postgres)` CI job (added in #126) will exercise the e2e tests on every PR.

## Test plan

- [ ] CI: stardag-api SQLite pytest passes
- [ ] CI: stardag-api Postgres pytest passes (new job exercises this)
- [ ] Reviewer skim of the cast emission (CAST AS uuid for equality, `::text` for ILIKE)
- [ ] After merge: deploy via stardag-cloud submodule bump → re-test the production UI Task Explorer build_id filter

### DIFF
--- a/app/stardag-api/src/stardag_api/routes/search.py
+++ b/app/stardag-api/src/stardag_api/routes/search.py
@@ -131,12 +131,26 @@ def build_jsonb_condition(
             return f"{task_alias}.{key} ILIKE '%' || :{param_name} || '%'", False, None
         return f"{task_alias}.{key} {sql_op} :{param_name}", False, None
 
-    # Build fields - require join to events and builds tables
+    # Build fields - require join to events and builds tables.
+    # builds.id is a Postgres UUID column. Bound parameters arrive from the
+    # query string as text and asyncpg sends them as character varying; an
+    # implicit `uuid = varchar` comparison is rejected by Postgres
+    # ("operator does not exist: uuid = character varying"). Use explicit
+    # casts: column-to-text for substring match, parameter-to-uuid for
+    # equality so the PK index is still usable.
     if key == "build_id":
         param_name = f"filter_build_id{param_suffix}"
         if sql_op == "ILIKE":
-            return f"builds.id ILIKE '%' || :{param_name} || '%'", True, None
-        return f"builds.id {sql_op} :{param_name}", True, None
+            return (
+                f"builds.id::text ILIKE '%' || :{param_name} || '%'",
+                True,
+                None,
+            )
+        return (
+            f"builds.id {sql_op} CAST(:{param_name} AS uuid)",
+            True,
+            None,
+        )
 
     if key == "build_name":
         param_name = f"filter_build_name{param_suffix}"
--- a/app/stardag-api/src/stardag_api/routes/search.py
+++ b/app/stardag-api/src/stardag_api/routes/search.py
@@ -6,7 +6,7 @@ from collections import Counter
 from typing import Annotated, Any
 from uuid import UUID
 
-from fastapi import APIRouter, Depends, Query
+from fastapi import APIRouter, Depends, HTTPException, Query
 from sqlalchemy import func, select, text
 from sqlalchemy.ext.asyncio import AsyncSession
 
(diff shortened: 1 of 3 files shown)


andhus (commented):
## Review responses

Bundled the cheap, high-value items in commit 7f82ac1. Per-concern status:

### 1. Malformed UUID still 500s — RESOLVED

New ``_validate_filter_value`` helper in ``routes/search.py`` rejects non-UUID values for ``build_id`` for any operator except ``~`` (ILIKE substring match). Raises ``HTTPException(400)`` with a clear message. Hooked into ``parse_filter_string`` so every call site (search, search/values, etc.) gets it for free. The 400 now reads:

```
Invalid build_id filter value 'not-a-uuid': must be a valid UUID for non-substring comparisons
```

### 2. ``!=`` not exercised end-to-end — RESOLVED

``test_search_filter_by_build_id_inequality_excludes_match`` creates two builds and asserts ``filter=build_id:!=:<a>`` returns the other build's task and excludes the named one. Closes the e2e gap.

### 3. ``pg_client`` / ``client`` near-duplicate — DEFERRED

Fine as a separate refactor PR once we've landed more PG-only tests. The risk is concrete (auth-override drift) but small until there are 3+ PG tests; right now there are 2 distinct locations that would benefit.

### 4. ``dependency_overrides.clear()`` blast radius — DEFERRED

Pre-existing pattern in ``client``; not introduced here. Worth a sweep across both fixtures together.

### 5. Audit other UUID columns in filter paths — DONE

Grep confirmed: ``build_jsonb_condition`` is the single funnel for filter SQL emission and only ``build_id`` is a UUID column. All other filter targets are varchar (``task_name``, ``task_namespace``, the ``task_id`` hash on the tasks table, ``version``) or JSONB paths. No other routes need fixing for ``/tasks/search``.

### 6. Minor — RESOLVED

- Comment in ``build_jsonb_condition`` broadened to "non-substring comparisons" — the branch correctly covers ``>``, ``<``, ``>=``, ``<=`` too (Postgres supports these between uuids).
- ``test_build_id_inequality_casts_param_to_uuid`` now asserts ``needs_join is True`` like its peers.
- Added a comment to the ILIKE test about why ``prefix = build_id[:8]`` is OK for the single-build scenario (UUID7 timestamp prefix collision risk is a no-op when only one build exists in the test).

### New tests added

- ``TestBuildIdFilterValidation`` (6 unit tests) — exercise the validation helper directly. Run on every backend, no DB required.
- ``test_search_filter_by_build_id_inequality_excludes_match`` (PG e2e) — covers ``!=``.
- ``test_search_filter_build_id_with_malformed_uuid_returns_400`` (PG e2e) — confirms the 400 propagates through FastAPI.
- ``test_search_filter_build_id_ilike_accepts_non_uuid_substring`` (PG e2e) — confirms the ``~`` carve-out keeps any string.

### Verification

- pytest SQLite: 216 passed, 12 skipped (PG-only).
- pytest with ``STARDAG_API_TEST_DATABASE_URL``: 227 passed, 1 skipped.
- pre-commit clean (ruff, pyright, prettier).

Ready to merge alongside Tier 3 cloud submodule bump.

PR #129, opened 2026-04-28
Fix UI logout-on-deploy: optional stable JWT_PRIVATE_KEY from Secrets Manager

andhus:
## Summary

Fixes the UI session loss across every API deploy. Diagnosis from prod investigation:

When ``JWT_PRIVATE_KEY`` is unset, ``InternalTokenManager`` generates a fresh ephemeral RSA keypair on every container start (``auth/tokens.py:107``). The UI exchanges its OIDC token for an internal JWT signed by this keypair and caches it in ``localStorage`` keyed by its 10-min ``expiresAt``. After every deploy:

- New container has a new keypair.
- UI's cached internal token (signed by the OLD key) fails signature verification on the new pod → 401.
- Cache thinks the token is still fresh (``expiresAt > now+30s``), so the UI never refreshes.
- Result: workspace dropdown populated (Cognito ID token still valid) but environments / builds / search all silently 401 → empty state with no error message until the user manually clears site data and re-logs in.

## Fix

Wire ``JWT_PRIVATE_KEY`` as an optional ECS task secret in ``api-stack.ts``, sourced from AWS Secrets Manager when ``STARDAG_API_JWT_PRIVATE_KEY_SECRET_NAME`` is set at synth time. The secret is expected to be a JSON document with a ``private_key`` field containing the PEM-encoded RSA private key (idiomatic AWS Secrets Manager pattern).

Default OSS path (env var unset) preserves the ephemeral keypair behaviour, which is fine for fresh deploys and single-task local setups.

## Production rollout

After this PR merges, the matching ``stardag-cloud`` PR will:

1. Generate an RSA-2048 keypair (one-time):
   ```bash
   openssl genrsa -out jwt-private.pem 2048
   ```
2. Provision in Secrets Manager:
   ```bash
   aws secretsmanager create-secret \
     --name stardag/api/jwt-private-key \
     --secret-string "{\"private_key\": $(jq -Rs . < jwt-private.pem)}"
   ```
3. Set ``STARDAG_API_JWT_PRIVATE_KEY_SECRET_NAME=stardag/api/jwt-private-key`` in ``cloud/config/.env.deploy``.
4. Bump submodule pointer.
5. Merge → deploy.

The first task to come up signs new tokens with the stable key; subsequent deploys reuse the same key so existing UI sessions continue to validate. **No more deploy-induced re-logins.**

## Not in scope (followup)

Defense-in-depth UI 401 refresh + session-expired UX is tracked in a separate ``stardag-ui`` PR. Even with this fix, normal 10-min internal-token TTL expiry can still leave the UI in a silent empty-state if the user is idle past 10 min — the UI fix handles 401 → re-exchange OIDC for a fresh internal token → retry, or surface "session expired" with a clear re-login button if Cognito itself has timed out.

## Verification

- ``npx cdk synth StardagApi`` with env var set → renders:
  ```
  - Name: JWT_PRIVATE_KEY
    ValueFrom: arn:aws:secretsmanager:us-east-1:763997220528:secret:stardag/api/jwt-private-key:private_key::
  ```
  IAM grant added (`secretsmanager:GetSecretValue`).
- Synth without env var → no ``JWT_PRIVATE_KEY`` in template (OSS-friendly).
- ``npm run build`` + ``npm test`` (19 jest tests): clean.
- pre-commit clean.

## Test plan

- [ ] CI: CDK build + jest passes.
- [ ] Reviewer skim of the optional secret-mount pattern.
- [ ] After merge: provision the prod secret + matching stardag-cloud submodule bump PR.
- [ ] Post-deploy: redeploy a second time within an hour, confirm any UI session opened pre-redeploy still works without re-login.

### DIFF
--- a/infra/aws-cdk/lib/api-stack.ts
+++ b/infra/aws-cdk/lib/api-stack.ts
@@ -61,7 +61,12 @@ export class ApiStack extends cdk.Stack {
     // every existing UI session on every redeploy.
     // The secret is expected to be a JSON document with a "private_key"
     // field containing a PEM-encoded RSA private key.
-    const jwtPrivateKeySecretName = process.env.STARDAG_API_JWT_PRIVATE_KEY_SECRET_NAME;
+    // Trim whitespace so a typo like ``FOO=" "`` in an env file is
+    // treated as unset instead of silently turning into a broken
+    // ``fromSecretNameV2`` lookup at synth time.
+    const rawJwtSecretName =
+      process.env.STARDAG_API_JWT_PRIVATE_KEY_SECRET_NAME?.trim();
+    const jwtPrivateKeySecretName = rawJwtSecretName || undefined;
     if (apiAutoscaleMax < apiAutoscaleMin) {
       throw new Error(
         `STARDAG_API_AUTOSCALE_MAX (${apiAutoscaleMax}) must be >= ` +
@@ -141,7 +146,13 @@ export class ApiStack extends cdk.Stack {
 
     // Look up the optional JWT private key secret so its ARN is in the
     // task definition. The secret value must already exist in Secrets
-    // Manager when the task starts; ECS fetches it at container launch.
+    // Manager when the task starts; ECS fetches it at container launch
+    // using the *execution role* (not the task role). When the secret is
+    // referenced via ``ecs.Secret.fromSecretsManager`` in the container's
+    // ``secrets`` block below, CDK auto-grants ``GetSecretValue`` to the
+    // execution role; we add an explicit grant here for clarity and to
+    // ensure the policy is in place even if a future refactor changes
+    // how the secret is referenced.
     const jwtPrivateKeySecret = jwtPrivateKeySecretName
       ? secretsmanager.Secret.fromSecretNameV2(
           this,
@@ -150,7 +161,7 @@ export class ApiStack extends cdk.Stack {
         )
       : undefined;
     if (jwtPrivateKeySecret) {
-      jwtPrivateKeySecret.grantRead(taskDefinition.taskRole);
+      jwtPrivateKeySecret.grantRead(taskDefinition.obtainExecutionRole());
     }
 
     // Grant task role permission to send emails via SES
--- a/infra/aws-cdk/lib/api-stack.ts
+++ b/infra/aws-cdk/lib/api-stack.ts
@@ -53,6 +53,15 @@ export class ApiStack extends cdk.Stack {
     // initial desired count.
     const apiAutoscaleMin = intEnv("STARDAG_API_AUTOSCALE_MIN", apiDesiredCount);
     const apiAutoscaleMax = intEnv("STARDAG_API_AUTOSCALE_MAX", 4);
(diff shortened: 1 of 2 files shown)


andhus (commented):
## Copilot review responses

All 3 concerns addressed in commit c3f0542. Per-comment status:

### 1. ``STARDAG_API_JWT_PRIVATE_KEY_SECRET_NAME`` not validated — RESOLVED

Added ``trim()`` on the env var value and treat empty / whitespace-only strings as unset. A typo like ``FOO=" "`` in an env file now falls back to the OSS-friendly ephemeral-keypair path instead of synthing a broken ``fromSecretNameV2`` lookup.

```ts
const rawJwtSecretName = process.env.STARDAG_API_JWT_PRIVATE_KEY_SECRET_NAME?.trim();
const jwtPrivateKeySecretName = rawJwtSecretName || undefined;
```

### 2. New conditional behaviour not covered by jest — RESOLVED

Added ``ApiStack JWT_PRIVATE_KEY secret wiring`` suite (4 new tests):

- **env var unset** → no ``JWT_PRIVATE_KEY`` secret in the API task def
- **env var set** → task def includes the Secrets Manager reference for the ``private_key`` JSON field. Matcher walks the ``ContainerDefinitions[].Secrets`` array and asserts the assembled ``Fn::Join`` ARN contains both the secret name and the ``:private_key::`` suffix
- **env var set** → IAM policy attached to the execution role grants ``secretsmanager:GetSecretValue`` on the secret ARN
- **env var = whitespace** → treated as unset (no JWT_PRIVATE_KEY in template)

The tests use a small ``synthApiStack(jwtSecretName)`` helper that flips the env var around CDK synth and restores it on teardown, so they don't bleed state into other tests. Existing 19 ``StardagStack`` tests still pass; total 23 jest tests.

### 3. (Correctness) Wrong IAM role for secret grant — RESOLVED

Good catch. ``ecs.Secret.fromSecretsManager(...)`` is resolved by the ECS agent at task launch using the **execution role**, not the task role. Granting only the task role wouldn't allow the task to start. Switched the explicit grant to:

```ts
jwtPrivateKeySecret.grantRead(taskDefinition.obtainExecutionRole());
```

In practice CDK already auto-grants the execution role when a secret is referenced in the container's ``secrets`` block, so a missing explicit grant here would have worked anyway — but the explicit grant should target the role that actually needs the permission, both for clarity and as belt-and-braces against future refactors. The new jest test pins this by asserting the IAM policy is attached to a role whose logical ID contains ``ExecutionRole``.

**Did not touch** the pre-existing DB secret grants on ``taskDefinition.taskRole`` (lines 144-145). They follow the same pattern but are inherited from the original commit and out of scope for this PR — happy to do a separate cleanup PR if you want, but it'd be a no-op behaviour-wise (the DB secrets are also referenced via ``ecs.Secret.fromSecretsManager`` in the ``secrets`` block, so CDK's auto-grant on the execution role keeps them working).

### Verification

- ``npm run build`` clean
- ``npm test`` → 23 passed (19 existing + 4 new)
- pre-commit clean
- Smoke-tested ``cdk synth`` with env var set/unset/whitespace; behaviour matches assertions

PR #133, opened 2026-04-29
Discover-time task registration + bulk register + UX polish

andhus:
## Summary

End-to-end overhaul of how tasks reach the registry during a build, motivated by "tasks don't appear in the UI in the order they're discovered, sometimes only after they finish executing." Multiple logical changes kept in one PR because they share scaffolding and the tests cross-reference each other.

### 1. Tasks are registered at discover-time, not at start-time
The build engine (sync + async, sequential + concurrent) used to send `task_register` only when each task became runnable, so leaves arrived in the registry first and roots last. Now every task discovered during the walk is registered before any task starts running, so the full DAG appears in the UI immediately.

### 2. Post-order discovery walk
The walk recurses into deps first and registers the parent only after every child has registered. This means a parent's `dependency_task_ids` always resolve to existing rows in `_reconcile_dependency_edges` — eliminating the brief phantom-row window where the UI used to flash up `tid[:12]`-style placeholder names between parent and child registration. Same trick on the dynamic-dep path: `discover(dep)` runs before `task_add_dependencies_aio` so the dep row exists before the edge insert.

The concurrent build's discover walk was also vulnerable to a diamond-DAG race: two siblings' `discover()` calls for a shared dep could see the fast-path `if task.id in task_states` and append their own parents to `pending_registrations` before the original discoverer appended the dep. Fixed via per-task `discover_done` events that the fast-path now `await`s.

### 3. Bulk-register endpoint
New `POST /api/v1/builds/{id}/tasks/bulk` registers up to 1000 tasks in one transaction, processing the array in order so within-batch dep references resolve to existing rows (no phantom-creation in the happy path). Deduplicates by `task_id`, keeping the first occurrence. Same `TASK_PENDING` / `TASK_REFERENCED` event semantics as the single-task endpoint.

**Inside the transaction it's now ~6 SQL round-trips constant in batch size**, not the ~5N the original implementation had: pre-query existing tasks → sorted `SELECT FOR UPDATE` on existing rows only (deadlock mitigation) → bulk INSERT new tasks (UUID7 client-side PKs, single `executemany`) → bulk INSERT edges (computed in-Python from `pk_by_task_id`) → bulk INSERT events with microsecond-spaced `created_at` for stable ordering → in-Python `apply_event_to_task` on ORM rows → commit.

Optional `?id_only=true` query param returns only `{id, task_id}` per task instead of the full `TaskResponse` (~10× smaller response). The SDK passes this since it discards the response.

### 4. SDK adopts bulk register, with chunking + gzip + fallback
Build engine accumulates discovered tasks into `pending_registrations` (post-order) and flushes via one `task_register_bulk_aio` call after each discover walk (initial + each dynamic-deps yield). Chunked at **50 tasks per HTTP call** — well under the API's 1000 hard cap so DB transactions stay short, request bodies stay friendly even with fat task specs, and a chunk failure in `warn` mode loses few tasks before per-task fallback. For a 5000-task DAG this is 100 bulk POSTs instead of 5000 individual ones.

JSON request bodies above 1 KB are **gzipped on the wire** (typical bulk payloads compress 5–10× from repeated structure). The server's new `GZipRequestMiddleware` decompresses transparently — pure ASGI, **streaming** with running caps on both compressed (10 MB) and decompressed (50 MB) size so gzip-bombs can't allocate unbounded memory. Old SDKs and non-gzipped requests pass through unchanged.

Bulk-register falls back to per-task `task_register_aio` on a "missing route" 404 (older API deployments). The fallback is gated on the FastAPI default `{"detail": "Not Found"}` body so it doesn't trigger on app-level 404s like `"Build not found"`.

### 5. Stable per-build ordering in the API
`GET /builds/{id}/tasks` now joins against `events` filtered to this build and orders by `min(events.created_at), Task.id` — so a re-referenced task appears at the position where it was *seen in this build*, not where it was first ever inserted in the environment. Postgres-safe (no `min(uuid)` aggregate).

### 6. UX polish: phantom rows hidden from list views
`is_phantom` is now exposed on `TaskResponse` / `TaskWithStatusResponse`. The UI's `BuildView` filters phantoms out of the table list and the "X tasks" counter, so even if a phantom slips through (build crash mid-discover, out-of-band register, etc.) the user doesn't see ugly placeholder names alongside real tasks. The DAG view still renders them — dropping nodes there would break edges.

### 7. Robustness: `build_fail` on discovery error, warn-mode resilience
- If a task's `requires()` / `complete()` raises during discovery, `build_fail[_aio]` is now emitted rather than the build being left RUNNING forever.
- `task_start[_aio]` / `task_complete[_aio]` / `task_resume_aio` all wrapped in `handle_registry_error` so `warn` mode genuinely tolerates registry hiccups (without hard-failing on a 404 from `/start` after a discover-time register failure).

## Behaviour changes to be aware of

- **`APIRegistry.task_start[_aio]` no longer auto-calls `task_register[_aio]`.** Contract is now "register first" everywhere — `/start` is a pure event endpoint. Internal callers (build engines, Prefect integration) updated. External callers that used `task_start_aio` directly without a prior `task_register_aio` will now hit a 404 from `/start`. Migration: add an explicit `await registry.task_register_aio(build_id, task)` before `task_start_aio`.
- **Sequential build registers in post-order DFS** (deps before parents) where it previously used pre-order. Concurrent build is approximately post-order — siblings still interleave but every task's deps are guaranteed to register before it. Visible only in registry / UI ordering; no API breakage.
- **`_reconcile_dependency_edges` phantom-creation is now a safety hatch**, not the happy path. Documented inline. Triggers only on edge cases like build crash mid-discover or out-of-band edge registrations.

## Backwards compatibility

- **Old SDK ↔ new server**: works unchanged. Per-task `POST /tasks` endpoint hasn't changed semantics; new `is_phantom` field is silently ignored by old Pydantic models.
- **New SDK ↔ old server**: works. `/tasks/bulk` 404s → SDK falls back to per-task register (no perf win until API is upgraded, but builds keep working).
- **Server-first deploy recommended** so early SDK upgraders immediately benefit from bulk register, but either order works.

## Test plan

- [x] `pytest lib/stardag/tests/test_build/` — 118 passed
- [x] `pytest lib/stardag/tests/test_registry/` — 56 passed
- [x] Full SDK suite — 503 passed (excluding modal which needs cloud creds)
- [x] `pytest app/stardag-api/tests/` — 250 passed, 17 skipped
- [x] All 9 CI checks green: Integration Tests, Integration Tests (Browser), Lint, Test Frontend, Test Python 3.11/3.12/3.13/3.14, Test Python (stardag-api on Postgres) — the Postgres job exercises real `FOR UPDATE` lock acquisition and the deadlock-mitigation test
- [x] `pre-commit` clean (ruff, ruff-format, pyright × 2, eslint, tsc)
- [ ] Post-deploy: canary deploy first, watch p99 of `/tasks/bulk` for one hour, verify in a real build that the full DAG renders before any task starts and that the table view stays stable across refreshes.

## New tests (29)

**SDK (`lib/stardag/tests/`)**:
- `test_build/test_sequential.py`:
  - `TestDiscoverTimeRegistrationSequential`, `TestDiscoverTimeRegistrationAio` — registration-before-start invariant, exactly-once-per-build, dynamic deps registered when discovered.
  - `TestDiscoverTimeRegistrationErrorHandling` — `warn` mode tolerates `/start` 404; `raise` propagates; concurrent build skips `/start` when registration never landed.
  - `TestDiscoveryFailureBuildFail` — `build_fail[_aio]` emitted when a task's `requires()` raises.
  - `TestBulkRegister` — exactly-1 batch for static DAG, post-order, dynamic deps trigger separate batch.
  - `TestNoPhantomsHappyPath` — bulk array always orders deps before parents (diamond DAG).
  - `TestConcurrentDiscoverRaceFreedom` — diamond-DAG race fix: shared dep before *both* parents.
  - `TestBulkRegisterChunking` — `monkeypatch`ed chunk size of 3 → 7-task DAG produces three batches of [3, 3, 1] preserving post-order.
- `test_registry/test_api_registry.py`:
  - `TestMaybeGzipJsonBody` — None / small / threshold-exact / large body gzip decision.
  - `TestAPIRegistryGzipsWireFormat` — `httpx.MockTransport`-based: 50 distinct fake tasks → outgoing request has `Content-Encoding: gzip` *and* `id_only=true` query param; small single-task → no gzip.

**API (`app/stardag-api/tests/`)**:
- `test_builds.py`:
  - `test_list_tasks_in_build_returns_stable_order` — stable insertion order across repeated calls.
  - `test_list_tasks_in_build_orders_by_per_build_first_event` — re-referenced task lands at its current-build position.
  - `test_bulk_register_creates_tasks_and_resolves_in_batch_deps` — in-batch dep resolution, no phantoms in happy path.
  - `test_bulk_register_upgrades_existing_phantoms` — phantom upgrade in place.
  - `test_bulk_register_empty_array_returns_201_empty`, `test_bulk_register_caps_batch_size`.
  - `test_bulk_register_dedupes_within_batch` — first-occurrence wins.
  - `test_bulk_register_referenced_event_for_existing_task`.
  - `test_bulk_register_creates_phantom_for_unknown_upstream` — explicit safety-hatch phantom test.
  - `test_bulk_register_array_order_maps_to_list_order` — bulk path's user-visible ordering contract.
  - `test_concurrent_bulk_registers_overlapping_tasks` — two parallel POSTs with reversed orders, both succeed (Postgres-meaningful).
  - `test_bulk_register_id_only_returns_slim_response` / `test_bulk_register_default_returns_full_response` — slim vs full response shapes.
- `test_gzip_request.py`:
  - Round-trip via test client, pass-through for non-gzipped, unknown-encoding pass-through, malformed gzip → 400, single-task gzip works.
  - `test_gzip_bomb_aborts_with_413` — streaming decompression aborts mid-stream when decompressed cap breached (`monkeypatch`ed to 1 KB).
  - `test_oversized_compressed_body_rejected` — compressed-size cap is the first line of defence.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,9 +33,17 @@ notes; the bullets below are the per-component summary.
   before the edge insert.
 - **Bulk register**: build engines now collapse the discovered tasks
   into a single `task_register_bulk(_aio)` call per discover walk
-  (initial + each dynamic-deps yield), chunked at 1000 tasks per HTTP
-  request. For large fan-out DAGs this is a dramatic reduction in HTTP
-  round-trips. Per-task fallback on 404 (older API deployments).
+  (initial + each dynamic-deps yield), chunked at 50 tasks per HTTP
+  request (well under the API's 1000 hard cap so DB transactions stay
+  short and request bodies stay friendly even with fat task specs).
+  For large fan-out DAGs this is a dramatic reduction in HTTP
+  round-trips — a 5000-task DAG goes from 5000 individual POSTs to
+  100 bulk POSTs. Per-task fallback on 404 (older API deployments).
+- **Gzipped request bodies on the wire**: JSON request bodies above 1KB
+  are gzipped client-side before sending; bulk-register payloads with
+  repeated structure compress 5–10× typically. The server's new
+  `GZipRequestMiddleware` decompresses transparently — old SDKs and
+  non-gzipped requests pass through unchanged.
 - **`build_fail(_aio)` now emitted on discovery error**: if a task's
   `requires()` / `complete()` raises during discovery, the registry
   receives `build_fail` rather than the build being left RUNNING
@@ -76,6 +84,12 @@ notes; the bullets below are the per-component summary.
   phantom-creation in `_reconcile_dependency_edges`). Deduplicates by
   `task_id`, keeping the first occurrence. Same TASK_PENDING /
   TASK_REFERENCED event semantics as the single-task endpoint.
+- **`GZipRequestMiddleware`**: ASGI middleware that decompresses
+  incoming `Content-Encoding: gzip` request bodies before route
+  handlers parse them. Pass-through for non-gzipped requests so old
+  SDK versions, direct `curl` callers, and non-bulk endpoints keep
+  working unchanged. Returns 400 on malformed gzip so clients see a
+  clear error rather than a downstream parse failure.
 - **`is_phantom` on `TaskResponse` / `TaskWithStatusResponse`**: the
   flag has existed on the `Task` model for a while; it's now exposed
   in the response so the UI (and other consumers) can distinguish
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -84,6 +84,9 @@ notes; the bullets below are the per-component summary.
   phantom-creation in `_reconcile_dependency_edges`). Deduplicates by
   `task_id`, keeping the first occurrence. Same TASK_PENDING /
   TASK_REFERENCED event semantics as the single-task endpoint.
+  Optional `?id_only=true` query param returns only `{id, task_id}`
+  per task instead of the full `TaskResponse` (~10× smaller
+  response — the SDK passes this since it doesn't read the response).
 - **`GZipRequestMiddleware`**: ASGI middleware that decompresses
   incoming `Content-Encoding: gzip` request bodies before route
   handlers parse them. Pass-through for non-gzipped requests so old
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,104 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
 ## [Unreleased]
 
+## [0.6.0] — Discover-time task registration + bulk register endpoint
+
+End-to-end overhaul of how tasks reach the registry during a build,
+motivated by "tasks don't appear in the UI in the order they're
+discovered, sometimes only after they finish executing." See
+[RELEASE_NOTES.md](RELEASE_NOTES.md#v060) for the full story and migration
+notes; the bullets below are the per-component summary.
+([#133](https://github.com/stardag-dev/stardag/pull/133))
+
+### SDK
+
+#### New behaviour
+
+- **Discover-time registration**: every task discovered during the build
+  walk is now registered with the registry _before_ any task starts
+  executing. The full DAG appears in the UI immediately rather than
+  leaves-first as tasks become runnable. Applies to both `build` /
+  `build_aio` and `build_sequential` / `build_sequential_aio`.
+- **Post-order discovery walk**: `discover()` recurses into static deps
+  first and only registers the parent once every child has registered.
+  Eliminates the brief phantom-row window where the UI used to flash up
+  `tid[:12]`-style placeholder names between parent and child
+  registration. Same trick on the dynamic-dep path: `discover(dep)`
+  runs before `task_add_dependencies(_aio)` so the dep row exists
+  before the edge insert.
+- **Bulk register**: build engines now collapse the discovered tasks
+  into a single `task_register_bulk(_aio)` call per discover walk
+  (initial + each dynamic-deps yield), chunked at 1000 tasks per HTTP
+  request. For large fan-out DAGs this is a dramatic reduction in HTTP
+  round-trips. Per-task fallback on 404 (older API deployments).
+- **`build_fail(_aio)` now emitted on discovery error**: if a task's
+  `requires()` / `complete()` raises during discovery, the registry
+  receives `build_fail` rather than the build being left RUNNING
+  forever.
+
+#### Breaking changes
+
+- **`APIRegistry.task_start[_aio]` no longer auto-calls
+  `task_register[_aio]`.** The contract is now "register first"
+  everywhere — `/start` is a pure event endpoint. Internal callers
+  (build engines, Prefect integration) are updated. External callers
+  that used `task_start_aio` directly without first calling
+  `task_register_aio` will now hit a 404 from `/start`. Migration:
+  add the explicit `await registry.task_register_aio(build_id, task)`
+  call before `task_start_aio`.
+- **Sequential build registers tasks in post-order DFS** (deps before
+  parents) where it previously used pre-order. Concurrent build is
+  approximately post-order — siblings still interleave but every
+  task's deps are guaranteed to register before it. Visible via the
+  UI / registry only; no API breakage.
+
+#### Compatibility
+
+- **Backwards compatible with the old Registry API**: the SDK's
+  `task_register_bulk(_aio)` catches the FastAPI "missing route" 404
+  and falls back to per-task `task_register(_aio)`, mirroring the
+  pattern from `task_add_dependencies`. Existing
+  `task_register(_aio)`, `task_start(_aio)`, etc. endpoints are
+  unchanged.
+
+### API
+
+#### New features
+
+- **New endpoint `POST /builds/{build_id}/tasks/bulk`**: registers up
+  to 1000 tasks in a single transaction, processing the array in order
+  so within-batch dep references resolve to existing rows (no
+  phantom-creation in `_reconcile_dependency_edges`). Deduplicates by
+  `task_id`, keeping the first occurrence. Same TASK_PENDING /
+  TASK_REFERENCED event semantics as the single-task endpoint.
+- **`is_phantom` on `TaskResponse` / `TaskWithStatusResponse`**: the
+  flag has existed on the `Task` model for a while; it's now exposed
+  in the response so the UI (and other consumers) can distinguish
+  placeholder rows from real registered tasks.
+- **`GET /builds/{id}/tasks` orders by per-build first event**:
+  joins against `events` filtered to this build and orders by
+  `min(events.created_at), Task.id`. Re-referenced tasks now appear
+  at the position where they were _first seen in this build_, not
+  where they were first ever inserted in the environment. Previously
+  unordered (DB insertion order, effectively arbitrary).
+
+#### Behaviour change
+
+- **Phantom-creation in `_reconcile_dependency_edges` is now a safety
+  hatch**: with the SDK's post-order discover walk + the bulk
+  endpoint's in-array ordering, every dep `task_id` resolves to an
+  existing row in normal operation. Phantom-creation only triggers
+  when a build crashes mid-discover, when an out-of-band caller
+  registers an edge before its upstream task, or when an older SDK is
+  used. Documented inline.
+
+### UI
+
+- **Phantom rows hidden from the build's task table** and the "X tasks"
+  counter. They still render in the DAG view (dropping nodes there
+  would leave dangling edges). Reads `is_phantom` from the API's
+  `TaskResponse`.
+
 ## `app/stardag-api|ui` only — 2026-04-28
 
 > App (API + UI) changes only — no SDK release. Deployed continuously
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,114 @@ For changes to the Registry API, UI, and other components, see [CHANGELOG.md](CH
 
 ---
 
+## v0.6.0 — Discover-time task registration & bulk register
+
+The build engine now registers every task with the registry as it's
+discovered (in post-order DFS), in a single bulk HTTP call per discover
+walk. The full DAG appears in the UI immediately rather than appearing
+leaves-first as tasks become runnable, and the table view stays in a
+stable order across refreshes.
+
+### What changed in your build runs
+
+Before v0.6.0:
+
+```
+[Discovery walks the DAG locally]
+[Build runs]
+   leaf-1 starts → registry sees leaf-1 for the first time
+   leaf-2 starts → registry sees leaf-2 for the first time
+   ...
+   parent starts → registry sees parent
+[UI shows tasks appearing in execution order, often leaves-first]
+```
+
+In v0.6.0:
+
+```
+[Discovery walks the DAG locally — collecting tasks in post-order]
+[ONE HTTP call: POST /builds/{id}/tasks/bulk with [leaf-1, leaf-2, ..., parent]]
+[UI immediately shows the full DAG, in stable post-order]
+[Build runs — each task already exists in the registry]
+```
+
+For large fan-out DAGs (think: a `for chunk in chunks: yield Process(chunk)`
+yielding 500 chunks), this is the difference between 500 sequential
+HTTP round-trips and one bulk POST.
+
+### Compatibility — the short version
+
+- **Old SDK ↔ new server**: works unchanged. The per-task
+  `POST /builds/{id}/tasks` endpoint hasn't changed semantics, and the
+  new `is_phantom` field on responses is silently ignored by old
+  Pydantic models.
+- **New SDK ↔ old server**: works. The SDK calls the new bulk endpoint
+  first; on a 404 ("missing route") it falls back to per-task
+  registration with a warning. You lose the per-discover single-call
+  perf win until the API is upgraded, but builds keep working.
+- **You can deploy in either order.** Server-first is recommended so
+  early SDK upgraders immediately get the bulk-call optimisation.
+
+### Migration: direct callers of `APIRegistry.task_start[_aio]`
+
+> **Skip this section** if you only build via `build()`, `build_aio()`,
+> `build_sequential()`, or `build_sequential_aio()`. The build engines
+> are updated and shield you from the contract change.
+
+`APIRegistry.task_start[_aio]` no longer auto-calls `task_register[_aio]`
+before emitting the start event. The `/start` endpoint is now pure — it
+returns 404 if the task hasn't been registered.
+
+```python
+# v0.5.9 and earlier — task_start_aio auto-registered.
+await registry.task_start_aio(build_id, task)
+
+# v0.6.0 — you must register first.
+await registry.task_register_aio(build_id, task)
+await registry.task_start_aio(build_id, task)
+```
+
+If you subclass `RegistryABC` directly: no migration needed — your
+subclass continues to work as before. The default
+`RegistryABC.task_register_bulk[_aio]` falls back to looping
+`task_register[_aio]` per task, so you don't need to override it
+unless your backend has an efficient batch path.
+
+### Behaviour notes (most users won't notice)
+
+- **Sequential build registers in post-order DFS** (leaves before
+  parents) where it previously used pre-order. Visible only in the
+  registry / UI ordering, not in any API.
+- **`build_fail(_aio)` now emitted when discovery raises.** If a
+  task's `requires()` or `complete()` throws during discovery (rare),
+  the build is now properly marked failed in the registry instead of
+  being left in RUNNING state forever.
+- **Phantom rows are filtered out of the build task table in the UI.**
+  These auto-created placeholder rows used to flash up briefly between
+  parent and child registration; they no longer occur in normal
+  operation (post-order ensures deps register first), but if one
+  legitimately exists — e.g. a build crashed mid-discover and left
+  orphan rows — the UI hides it from the table. The DAG view still
+  renders it.
+
+### Also in this release
+
+- **New `RegistryABC.task_register_bulk[_aio]` method** with a default
+  loop-`task_register` implementation. `APIRegistry` overrides it to
+  POST `/tasks/bulk`. Custom registries get the loop default for free
+  but can override for batched backends.
+- **`is_phantom` exposed on `TaskResponse`** so other API consumers
+  can distinguish placeholder rows from real registered tasks.
+- **Discover-walk race fix in concurrent build**: when two siblings
+  shared a dep (diamond DAG), the fast-path
+  `if task.id in task_states: return` could let one sibling's parent
+  append to the registration order ahead of the still-in-flight dep,
+  re-introducing the phantom window. The fast-path now awaits a
+  per-task `discover_done` event, ensuring deps always register
+  before parents even under sibling concurrency.
+
+---
+
 ## v0.5.9 — Dynamic deps visible in the DAG view
 
 Dynamically-yielded dependencies now register as graph edges in the Registry,
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -32,14 +32,22 @@ In v0.6.0:
 
 ```
 [Discovery walks the DAG locally — collecting tasks in post-order]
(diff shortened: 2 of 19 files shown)


andhus:
Thanks for the review @copilot — addressed all 6 comments in cd437ee. Per-comment status:

**[builds.py L1274 — Task.created_at vs per-build ordering](https://github.com/stardag-dev/stardag/pull/133#discussion_r3163677909)** — fixed. `list_tasks_in_build` now joins against the `events` table filtered to this build and orders by `min(events.created_at), min(events.id)`. New API test `test_list_tasks_in_build_orders_by_per_build_first_event` asserts that a re-referenced task lands at its current-build position rather than at the top by virtue of a stale `Task.created_at`.

**[_sequential.py L505 — sync `task_start` unguarded after warn-mode register failure](https://github.com/stardag-dev/stardag/pull/133#discussion_r3163677971)** — fixed. `registry.task_start` (and `task_complete`) in `_run_task_sequential` are now wrapped in `try/except + handle_registry_error`. Same treatment as artifact upload below them. New test `test_warn_mode_tolerates_start_failure_sequential` covers the exact case (registry that 404s `/start`, build still succeeds locally in `warn` mode).

**[_sequential.py L983 — async `task_start_aio` unguarded, same issue](https://github.com/stardag-dev/stardag/pull/133#discussion_r3163678008)** — fixed. Same pattern in `_run_task_sequential_aio` for `task_start_aio` and `task_complete_aio`. Covered by `test_warn_mode_tolerates_start_failure_aio` parametrised across `build_aio` + `build_sequential_aio`.

**[_concurrent.py L997 — failed-retry register still proceeds to `task_start_aio`](https://github.com/stardag-dev/stardag/pull/133#discussion_r3163678053)** — fixed by combining your suggestion with explicit error handling: `submit_with_lock` now skips `task_start_aio` entirely when `state.registered` is False (no point firing a known-to-404 call) AND wraps the call in `handle_registry_error` for the success path so transient blips can't escalate. `task_resume_aio` got the same protection. New test `test_warn_mode_skips_start_when_registration_failed_concurrent` uses a registry that fails every `task_register_aio` and asserts `task_start` is never called and the build still succeeds.

**[_sequential.py L200 — discovery before `build_fail` try/except](https://github.com/stardag-dev/stardag/pull/133#discussion_r3163678094)** — fixed. Discovery + previously-completed marking are now inside the existing try/except, so a thrown `requires()` / `complete()` reaches the existing `build_fail` handler. New test `test_sequential_emits_build_fail_when_discover_raises` uses a task whose `requires()` raises and asserts the registry sees `build_start` then `build_fail` (and never `build_complete`).

**[_sequential.py L712 — same issue in async sequential variant](https://github.com/stardag-dev/stardag/pull/133#discussion_r3163678119)** — fixed by the same restructure. Covered by `test_aio_emits_build_fail_when_discover_raises` parametrised across `build_aio` + `build_sequential_aio` (catches the `RuntimeError` directly for sequential and inside the `TaskGroup`'s `ExceptionGroup` for concurrent).

**Suppressed comment (concurrent build, same as 5/6)** — the discovery + `mark_previously_completed` + `task_executor.setup()` block in `build_aio` is now inside the `try:` whose `except` calls `build_fail_aio`, with the existing `finally: task_executor.teardown()` (which is robust to a never-set-up executor) handling cleanup. Same `test_aio_emits_build_fail_when_discover_raises[concurrent]` covers it.

All tests still pass: 109 build tests, 232/249 API tests (17 skipped due to env), and pre-commit (ruff, ruff-format, pyright × 2) is clean.

andhus:
Layered on the post-order + bulk-register improvements per the follow-up discussion. New commit: 7bf563a.

**What's added**

1. **Post-order discover** in both build flavors (`_concurrent.py`, `_sequential.py`). Recurse into deps first, register self only after all children have registered. With this, a parent's `dependency_task_ids` always resolve to existing rows in `_reconcile_dependency_edges` — no phantom-creation in the happy path. Eliminates the ~ms-to-s phantom window where the UI used to flash up `tid[:12]`-style placeholder names between parent and child registration.

2. **Swap dynamic-dep order**: in `process_result` (concurrent) and `_run_task_sequential[_aio]` (sequential), `discover(dep)` now runs *before* `task_add_dependencies_aio(...)` so the dep row exists before the edge insert reaches the API.

3. **Bulk register API endpoint** `POST /api/v1/builds/{id}/tasks/bulk` — single transaction, processes the array in order so within-batch dep references resolve. Capped at 1000 tasks per call. Reuses phantom-upgrade logic. Limit checks: rate-limited as 1 request, entity-creation limits applied per item.

4. **SDK adopts bulk register** with per-task fallback on 404 (older API deployments). Build engine accumulates discovered tasks into `pending_registrations` (post-order) and flushes via one `task_register_bulk_aio` call after each discover walk (initial + each dynamic-deps yield). Dramatic reduction in HTTP round-trips for large fan-out DAGs.

5. **Phantom-path documented as a safety hatch** in `_reconcile_dependency_edges`. Phantoms still trigger if a build crashes mid-discovery or if an out-of-band caller registers an edge before the upstream task — so it stays as defense-in-depth, but the docstring now flags it as "should not normally trigger".

6. **`is_phantom` exposed** in `TaskResponse`/`TaskWithStatusResponse`. UI's `BuildView` filters phantoms out of the table list and the "X tasks" counter (they'd show as ugly placeholder names if any leak through). DAG view still renders them — dropping nodes would break edges; visual-distinction styling is a nice-to-have follow-up.

**New tests (10)**

- API: bulk endpoint creates tasks with in-batch dep resolution, no phantoms in happy path
- API: bulk endpoint upgrades existing phantoms in place
- API: empty bulk returns 201 + empty array
- API: bulk over cap returns 400
- API: bulk emits `task_referenced` for an existing task
- SDK: sequential bulk emits exactly 1 batch for a static DAG
- SDK: sequential bulk preserves post-order
- SDK: async (concurrent + sequential) bulk emits 1 batch with parents-after-deps invariant
- SDK: dynamic deps trigger a separate bulk call (so the dep row exists before the edge insert)
- SDK: bulk array always orders deps before parents (verified for diamond DAG)

Existing test `test_registration_order_is_dfs_from_roots` was updated to expect post-order (was pre-order before this commit) and renamed accordingly.

**All tests green**
- 115 build tests
- 494 SDK tests overall (excluding modal which needs cloud creds)
- 32 build-API tests, 237 stardag-api tests overall
- pre-commit: ruff + ruff-format + pyright × 2 + eslint + tsc — all pass

CI re-running on the new commit now.

andhus:
Thanks for the thorough review @copilot — addressed all 10 comments in 7472ac0. Per-comment status:

**[_sequential.py L251 — chunk sequential bulk to ≤1000](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166560649)** / **[_concurrent.py L645 — same in concurrent](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166560746)** / **[_sequential.py L778 — same in async sequential](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166560893)** — fixed. Added `_BULK_REGISTER_CHUNK_SIZE = 1000` constants in both `_concurrent.py` and `_sequential.py` and rewrote `flush_pending_registrations[_aio]` to chunk in order, with per-chunk `handle_registry_error` and stop-on-first-failure semantics (preserves the warn-mode "successful chunks stay registered, failed chunks fall back to per-task retry" property). New tests `TestBulkRegisterChunking::test_sequential_chunks_large_batch` and `…test_concurrent_chunks_large_batch` use `monkeypatch` to lower the chunk size to 3 and verify a 7-task DAG produces three batches of [3, 3, 1] with post-order preserved across chunks.

**[schemas.py L155 — docstring claims dedup but code didn't](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166560785)** / **[builds.py L1049 — implement dedup or 400](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166560986)** — fixed by *implementing* dedup (first-occurrence wins). The schema docstring now matches the actual contract. Bulk endpoint deduplicates `tasks_in` upfront on `task_id`, preserving input order. New test `test_bulk_register_dedupes_within_batch` sends `[dupe-x(first), dupe-x(second), other-y]` and asserts only two responses come back, the first occurrence's `task_namespace`/`task_data` win, and exactly two `task_pending` events are emitted (not three).

**[_api_registry.py L437 — sync client should validate ≤1000](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166560836)** / **[_api_registry.py L803 — async same](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166561025)** — fixed. Added `_MAX_BULK_REGISTER_TASKS = 1000` in `_api_registry.py`. Both `task_register_bulk` and `task_register_bulk_aio` now raise `ValueError` with a clear message if a caller passes more than the cap. The build engine chunks above this layer so the check is purely defensive for direct external callers of `APIRegistry`.

**[_concurrent.py L616 — diamond DAG race in fast-path](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166560851)** — real bug, fixed. Added a `discover_done: dict[UUID, asyncio.Event]` map. The fast-path `if task.id in task_states` branch now `await`s the per-task `discover_done` event before returning, so a sibling discoverer for a shared dep can't append its own parent to `pending_registrations` ahead of the in-flight dep. The append + `event.set()` is wrapped in `try/finally` so a discover() that raises still releases waiters (rather than deadlocking inside the outer TaskGroup's cancellation). New test `TestConcurrentDiscoverRaceFreedom::test_diamond_dag_orders_shared_dep_before_all_parents` builds `parent → mid_a, mid_b → leaf` and asserts `leaf` appears before *both* `mid_a` and `mid_b` in the bulk batch — the exact case the race used to break.

**[builds.py L1024 — entity-limit overcounts when tasks already exist](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166561051)** — fixed. The bulk endpoint now does a single `SELECT Task.task_id WHERE task_id IN (...)` after the build/env check to find which task_ids already exist in the environment, then calls `check_entity_creation_limit("tasks", amount=new_task_count_estimate)` instead of `amount=len(tasks_in)`. A build that re-references all-cached tasks no longer burns the 24h tasks quota even though it creates 0 new Task rows. Events limit is unchanged (still `amount=len(tasks_in)` since each post-dedup task produces exactly one event).

**[test_builds.py L309 — docstring says "roots first" but it's post-order](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166561006)** — fixed. Test docstring now reads "static deps first, then their parents (post-order walk)".

**Tests + checks**
- 118 build tests (was 115; +3 new: chunking sequential + concurrent, diamond-DAG race)
- 33 build-API tests (was 32; +1 new: dedup)
- pre-commit clean (ruff, ruff-format, pyright × 2). Pyright caught a real possibly-unbound issue with `static_deps` in the refactored `discover()` — fixed by re-reading from `task_states[task.id].static_deps` after the lock release.

CI re-running on 7472ac0.

andhus:
Thanks for the security-focused review @copilot — addressed all 5 in bde6229.

**[gzip_request.py L78 — gzip-bomb / oversized-upload memory risk](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166994823)** — fixed. Rewrote the decompression path to be **streaming** with running caps on both the compressed input and the decompressed output. The middleware now uses `zlib.decompressobj(MAX_WBITS | 16)` with a per-chunk `max_length` budget, reads body chunks one-by-one from `receive`, and aborts the moment either limit is breached — without ever materialising the full decompressed body. Two limits enforced:
- `_MAX_COMPRESSED_BYTES = 10 MB` (first line of defence — bounded even for theoretical perfect-compression bombs).
- `_MAX_DECOMPRESSED_BYTES = 50 MB` (catches real-world bombs that inflate moderately past the compressed cap).

Two new tests in `tests/test_gzip_request.py` exercise this:
- `test_gzip_bomb_aborts_with_413` — `monkeypatch`es the decompressed cap to 1 KB, sends 100 KB of zeros (~100 B compressed) → 413 with "decompressed" in the detail.
- `test_oversized_compressed_body_rejected` — `monkeypatch`es the compressed cap to 256 B, sends 2 KB of `os.urandom` → 413 with "compressed" in the detail.

**[gzip_request.py L147 — `http.disconnect` after replayed body](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166994845)** — fixed. `replay_receive()` now returns an empty `http.request` with `more_body=False` on subsequent reads (a benign end-of-stream signal), instead of `http.disconnect` which can be interpreted as "client gave up" by defensive readers. The body was fully delivered, just already consumed — disconnect was a misleading signal.

**[test_api_registry.py L177 — `_Fake.id` is a class attribute](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166994865)** — fixed. Moved `self.id = uuid4()` into `__init__` so each fake task gets a distinct UUID. Removed the misleading "we override the UUID" comment (we don't, and now don't need to). The bulk gzip test now asserts `len(sent_ids) == 50` to prove the round-trip carries 50 unique IDs — exactly the realistic "many-distinct-tasks" payload the gzip benefit is supposed to demonstrate.

**[test_builds.py L444 — comment claims `TASK_PENDING` but it's `TASK_REFERENCED`](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166994888)** — fixed. The comment now correctly notes that phantom upgrade emits `TASK_REFERENCED` (because `task_already_existed=True` for the phantom row, even before upgrade) — not `TASK_PENDING`. The visibility claim still holds: `list_tasks_in_build` includes any task with any event in the build, so `TASK_REFERENCED` is enough.

**[test_builds.py L461 — name says 200 but asserts 201](https://github.com/stardag-dev/stardag/pull/133#discussion_r3166994924)** — fixed. Renamed `test_bulk_register_empty_array_returns_200_empty` → `…returns_201_empty` and updated the docstring to explain why 201 (the endpoint advertises `status_code=201`; the server doesn't bother distinguishing "0 created" from "N created").

**Tests + checks**
- 174 SDK lib tests (was 168; +2 streaming-bomb defence cases plus the strengthened bulk-distinct-IDs assertion)
- 40 build-API tests (was 38; +2 streaming-bomb tests in `test_gzip_request.py`)
- pre-commit clean (ruff, ruff-format, pyright × 2, eslint, tsc)

andhus:
Solid review — addressed in 7934a9c. The big lever (bulk-everything refactor) was worth doing now rather than a follow-up since the correctness items (deadlock, auth ordering) overlap with it.

## Performance — 5N round-trips → ~6 fixed

The per-task loop is gone. The bulk endpoint now does:

1. **One `SELECT` of existing tasks** (already there for the limit estimate; reused).
2. **One sorted `SELECT … FOR UPDATE`** scoped to existing rows only — new tasks aren't locked because there's no competing writer to serialize against (also closes the deadlock window — see below).
3. **One bulk `INSERT` of new Task rows** via `db.add_all` + a single `db.flush()`. UUID7 PKs are pre-generated client-side so SQLAlchemy/asyncpg batches them into one `executemany` round-trip instead of N. The PKs are stored in an in-Python `pk_by_task_id` map for downstream phases.
4. **One bulk `INSERT` of edges**, computed entirely in Python from `pk_by_task_id` + `dependency_task_ids`. The unknown-upstream phantom path stays as a separate batched SELECT/INSERT/SELECT — but only fires when an edge actually references a task that's neither in the batch nor in the DB (the safety hatch).
5. **One bulk `INSERT` of events** with explicit per-event `created_at = now + i µs` so the user-visible `min(events.created_at)` ordering on `list_tasks_in_build` still preserves array order. `apply_event_to_task` runs in Python on the in-memory Task ORM rows — no extra DB round-trip per task.
6. **One final `commit`**.

So a 1000-task batch is now ~6 SQL round-trips instead of ~5000.

## Correctness fixes

- **Lock ordering / deadlock** ([review point](https://github.com/stardag-dev/stardag/pull/133)) — fixed. The `SELECT … FOR UPDATE` runs once, scoped to existing rows only, **with `ORDER BY task_id`**. Two concurrent bulk POSTs into overlapping cached tasks now acquire locks in the same lexicographic order regardless of array order. Plus, since new tasks aren't locked at all, the bulk-of-bulk case (mostly fresh tasks) doesn't even touch the lock-ordering question. New test `test_concurrent_bulk_registers_overlapping_tasks` runs two parallel bulk POSTs with deliberately reversed array orders and asserts both succeed.
- **Auth/build check before limit checks** — fixed. Build/environment lookup moved ahead of `check_rate_limit` and `check_entity_creation_limit`. A probe call with a bogus `build_id` now returns 404 immediately without spending rate-limit budget or running the 24h entity-count query.
- **`new_task_count_estimate` divergence** — documented in the comment. Estimate runs without `FOR UPDATE`, so a concurrent transaction inserting the same `task_id` between the SELECT and the bulk INSERT can shrink the actual write count. This is strictly under-counting actual writes (never over-counting), so the limit guard rail stays safe — comment now says so explicitly.
- **Phantom upgrade ↔ `apply_event_to_task` invariant** — explicit comment in the upgrade block: we deliberately don't reset `latest_status_at` / `latest_status_event_id` because the immediately-following `apply_event_to_task(TASK_PENDING)` refreshes them via the `latest_status == PENDING` branch. If that branch ever gains a phantom-upgrade short-circuit, this needs an explicit reset.
- **404 fallback narrowness** — verified. `task_register_bulk_aio`'s fallback is gated on `_is_route_not_found` which only matches FastAPI's exact `{"detail": "Not Found"}` body. App-level 404s like `"Build not found"` carry a more specific detail string and re-raise normally — a transient missing-build wouldn't degrade to N per-task round-trips.
- **Within-batch event ordering on the wire** — explicit `created_at = now + i µs` per event keeps `list_tasks_in_build`'s `(min(events.created_at), Task.id)` ordering stable even though all events share the same flush. UUID7 was the wrong tiebreaker for re-referenced cached tasks (their Task.id comes from the original build); microsecond-spaced timestamps fix it. Comment in code ties the two together.
- **`apply_event_to_task` semantics for cached tasks** — already correct, no change needed (TASK_REFERENCED is a no-op, TASK_PENDING is a no-op on non-PENDING tasks). Confirmed by inspection.

## Test gaps

Three new tests in `app/stardag-api/tests/test_builds.py`:

- `test_bulk_register_creates_phantom_for_unknown_upstream` — explicit phantom-creation test for the safety hatch. Two child tasks both depending on an `orphan-up` that's not in the batch nor in the DB → asserts exactly one phantom row exists for `orphan-up` (not two — bulk reconcile must dedupe across the batch) and both child→phantom edges are present.
- `test_bulk_register_array_order_maps_to_list_order` — sends `[leaf-z, leaf-a, leaf-m, parent]` (deliberately non-alphabetical, parent last per post-order DFS) and asserts both the bulk-response array and the `GET /tasks` list endpoint preserve that order. Closes the gap that the bulk path's user-visible ordering contract was untested.
- `test_concurrent_bulk_registers_overlapping_tasks` — two parallel bulk POSTs into the same env with overlapping cached tasks and reversed array orders, both succeed. SQLite's `FOR UPDATE` is a no-op so this is mainly meaningful when CI runs against Postgres (the `Test Python (stardag-api on Postgres)` job exercises it).

## Defensive bits

- `_MAX_BULK_REGISTER_TASKS = 1000` comment now explicitly explains the 50/1000 asymmetry — 50 is the *working* size in the build engine, 1000 is the *defensive* ceiling for direct external `APIRegistry` callers. "Don't reduce this thinking 50 is the real limit — it isn't."

## Deferred

- **Lighter `TaskBulkResponse` with `id_only` query param** — defensible "small future improvement". Skipped for this PR; opening a follow-up ticket. The full TaskResponse round-trip is wasteful for the SDK-only use case but doesn't bite at chunk-size 50 (~50KB response, well within reason).

## Tests + checks

- 174 SDK tests, 248 API tests pass (was 174 + 245; +3 new bulk-endpoint tests).
- pre-commit clean (ruff, ruff-format, pyright × 2, eslint, tsc).

PR description: I'll update the perf framing in the description ("1 transaction, 1 round-trip" was sloppy; now it's "1 transaction, ~6 round-trips constant in batch size") in a separate edit.

PR #147, opened 2026-06-10
Fix compat_default hash-skip to compare raw value, not serialized

### LINKED ISSUE #146: compat_default hash-skip compares the *serialized* value, so it silently no-ops for enums / tuples / custom serializers
## Summary

`StardagField(compat_default=...)` is meant to keep a config that predates a
newly-added field hashing identically: in hash-mode serialization, a field whose
value equals its `compat_default` is dropped from the hash dump, so adding the
field with a backward-compatible default doesn't change existing hashes / task
identities.

But the drop comparison in `StardagBaseModel._handle_hash_mode` compares
`compat_default` against the **already-serialized** field value:

```python
data = handler(self)                      # fields serialized (mode="hash")
...
if stardag_field.compat_default == value or stardag_field.hash_exclude:
    continue                              # drop the field
```

For any field whose serialized form differs from its Python value — enums
(→ their `.value`), tuples (→ lists), or a field with a custom / hash-only
serializer — this comparison fails *even when the field is at its compat
default*, so the field is **not** dropped and the hash changes anyway. The
feature silently no-ops for those types.

There's also an asymmetry with the compat-*validation* path:
`_check_add_compatibility_defaults` injects `compat_default` as a **raw,
pre-validation** input (`data[name] = compat_default`). So the same
`compat_default` value is consumed in two different representations — raw on the
validate side, serialized on the hash side.

## Repro (stardag 0.7.3)

**Example A — enum-tuple field with a natural `compat_default`:**

```python
import enum
from typing import Annotated

import stardag as sdag
from stardag.base_model import StardagBaseModel


class Color(enum.StrEnum):
    RED = "red"
    GREEN = "green"


class Config(StardagBaseModel):
    colors: Annotated[
        tuple[Color, ...], sdag.StardagField(compat_default=(Color.RED,))
    ] = (Color.RED,)


print(Config().model_dump(mode="json", context={"mode": "hash"}))
# actual:   {'colors': ['red']}
# expected: {}   (field is at its compat default -> should be dropped)
```

To actually get the drop you have to declare the **serialized** form, which is
unintuitive (it isn't a natural "default value") and couples the declaration to
the serialization format:

```python
sdag.StardagField(compat_default=["red"])   # -> {} works, but it's a list of
                                            # strings, not a tuple of enums
```

**Example B — a field with a hash-only custom serializer:**

```python
from typing import Annotated, Any

from pydantic import WrapSerializer
import stardag as sdag
from stardag.base_model import StardagBaseModel, CONTEXT_MODE_KEY


def _hash_serialize(value: Any, handler, info):
    if info.context and info.context.get(CONTEXT_MODE_KEY) == "hash":
        return f"<{value}>"        # some canonical, hash-only string form
    return handler(value)


Canonical = Annotated[float, WrapSerializer(_hash_serialize)]


class Config(StardagBaseModel):
    weight: Annotated[Canonical, sdag.StardagField(compat_default=0.0)] = 0.0


print(Config().model_dump(mode="json", context={"mode": "hash"}))
# actual:   {'weight': '<0.0>'}
# expected: {}
```

Here there's no clean workaround: you'd have to hardcode the serializer's output
(`"<0.0>"`) as the `compat_default`, which (a) duplicates serialization logic,
(b) breaks if the serializer ever changes, and (c) is then *also* fed raw into
the compat-validate path, where it only round-trips if the serialized form
happens to be coercible back to the field type.

## Question / intended behavior

Is the serialized-value comparison intentional? It works transparently only for
fields whose serialized form equals their Python value (`bool`, `int`, `str`,
float-as-number, ...). For everything else it silently fails to drop the field,
which defeats the point of `compat_default` for those types.

If the intent is "drop when the field is at its compat default, regardless of
type", a couple of options:

1. Compare the **raw** field value rather than the serialized one — e.g.
   `getattr(self, name) == compat_default` — and drop from the serialized dict
   on match. This would also be symmetric with
   `_check_add_compatibility_defaults`, which uses the raw value.
2. Serialize `compat_default` with the same context/mode before comparing, so
   both sides are in the same space.

Happy to send a PR for whichever you prefer. If the current behavior *is*
intended, it would be worth documenting that `compat_default` must be supplied
in serialized form for any non-trivially-serialized field.

andhus:
## Summary

Fixes #146.

`StardagField(compat_default=...)` keeps a config that predates a newly-added
field hashing identically: in hash-mode serialization, a field whose value
equals its `compat_default` is dropped from the hash dump, so adding the field
with a backward-compatible default doesn't change existing task IDs/hashes.

The drop comparison in `StardagBaseModel._handle_hash_mode` compared
`compat_default` against the **already-serialized** field value:

```python
data = handler(self)                      # fields serialized (mode="hash")
...
if stardag_field.compat_default == value or stardag_field.hash_exclude:
    continue                              # drop the field
```

For any field whose serialized form differs from its Python value — enums
(→ `.value`), tuples (→ lists), or a field with a custom/hash-only serializer —
this comparison failed *even when the field was at its compat default*, so the
field was **not** dropped and the hash changed anyway. The feature silently
no-opped for those types.

## Fix

Compare the **raw Python value** (`getattr(self, name)`) against
`compat_default` (option 1 in the issue):

- Works transparently for all field types, regardless of serialized form.
- Symmetric with `_check_add_compatibility_defaults`, which already injects the
  raw `compat_default` on the validate side — so the same value is now consumed
  in one representation on both sides.
- `compat_default` is supplied in its natural Python form, not the
  serialization format.

Also guards the comparison with `compat_default is not _UNSET` so a sentinel
default is never matched.

## Tests

Added serialize cases for the two reported scenarios, each covering both
at-default (dropped in hash mode) and not-at-default (kept):

- enum-tuple field (`tuple[Color, ...]` serializing to a list of strings)
- field with a hash-only custom `WrapSerializer`

Both repros from the issue now produce `{}` as expected. Full unit suite passes;
ruff + pyright clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,20 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
 ## [Unreleased]
 
+### SDK
+
+- **`stardag/base_model.py`**: fixed `StardagField(compat_default=...)` so the
+  hash-mode drop compares the field's **raw Python value** against
+  `compat_default` instead of the already-serialized value. Previously the
+  feature silently no-opped for any field whose serialized form differs from
+  its Python value — enums (→ `.value`), tuples (→ lists), or fields with a
+  custom/hash-only serializer — so adding such a field with a compat default
+  still changed existing task IDs/hashes. The comparison is now symmetric with
+  the compat-validation path (which also uses the raw value), and
+  `compat_default` is given in its natural Python form rather than the
+  serialized form.
+  ([#146](https://github.com/stardag-dev/stardag/issues/146))
+
 ## [0.7.3] — 2026-05-08
 
 `stardag modal deploy` and `stardag modal stardag-api-key create` now
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,19 +6,29 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
 ## [Unreleased]
 
+## [0.8.0] — 2026-06-11
+
+Behaviour fix to `StardagField(compat_default=...)` that can change task
+IDs/hashes for fields with non-trivially-serialized types. See
+[RELEASE_NOTES.md](RELEASE_NOTES.md#v080--compat_default-compares-the-raw-python-value)
+for the migration guide.
+
 ### SDK
 
-- **`stardag/base_model.py`**: fixed `StardagField(compat_default=...)` so the
-  hash-mode drop compares the field's **raw Python value** against
-  `compat_default` instead of the already-serialized value. Previously the
-  feature silently no-opped for any field whose serialized form differs from
-  its Python value — enums (→ `.value`), tuples (→ lists), or fields with a
-  custom/hash-only serializer — so adding such a field with a compat default
-  still changed existing task IDs/hashes. The comparison is now symmetric with
-  the compat-validation path (which also uses the raw value), and
-  `compat_default` is given in its natural Python form rather than the
-  serialized form.
-  ([#146](https://github.com/stardag-dev/stardag/issues/146))
+- **`stardag/base_model.py`**: `StardagField(compat_default=...)`'s hash-mode
+  drop now compares the field's **raw Python value** against `compat_default`
+  instead of the already-serialized value. Previously the feature silently
+  no-opped for any field whose serialized form differs from its Python value —
+  enums (→ `.value`), tuples (→ lists), or fields with a custom/hash-only
+  serializer — so adding such a field with a compat default still changed
+  existing task IDs/hashes. The comparison is now symmetric with the
+  compat-validation path (which also uses the raw value): `compat_default` is
+  supplied in its natural validated Python form rather than the serialized
+  form. **Breaking** for the affected types — see the migration guide.
+  ([#146](https://github.com/stardag-dev/stardag/issues/146),
+  [#147](https://github.com/stardag-dev/stardag/pull/147))
+- Documented `StardagField.compat_default` / `hash_exclude` (previously a
+  `TODO` docstring).
 
 ## [0.7.3] — 2026-05-08
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,76 @@ For changes to the Registry API, UI, and other components, see [CHANGELOG.md](CH
 
 ---
 
+## v0.8.0 — `compat_default` compares the raw Python value
+
+Fixes `StardagField(compat_default=...)` so it works for all field types.
+The hash-mode drop now compares a field's **raw Python value** against
+`compat_default`, instead of the already-serialized value.
+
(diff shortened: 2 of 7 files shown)


andhus:
## Code Review

**Overview:** Focused, well-targeted fix. The hash-mode drop compared `compat_default` against the *already-serialized* value, so for any field whose serialized form differs from its Python value (enums → `.value`, tuples → lists, custom/hash-only serializers), the comparison failed even at the default — the field stayed in the dump and the hash changed. The fix compares the raw Python value (`getattr(self, name)`) and adds an `is not _UNSET` guard.

### Correctness — looks right ✅
- The core fix is sound and genuinely symmetric with `_check_add_compatibility_defaults`, which injects the raw `compat_default` on the validate side.
- `getattr(self, name)` is safe: it is only reached after `model_fields.get(name) is not None`. Since `model_fields` is keyed by field name (not serialization alias), `name` is guaranteed to be the real attribute name — no `AttributeError` risk and no regression for aliased fields.
- The `is not _UNSET` guard is a clarity/correctness improvement over relying on `object()` identity never matching.

### One subtlety worth a docstring note
The comparison is symmetric **only when `compat_default` is idempotent under validation** — i.e. given in the field’s natural validated Python form. If someone supplies `compat_default=[1, 2]` for a `tuple[int, ...]` field, the validate side coerces it to `(1, 2)` (stored value), but the serialize side compares `(1, 2) == [1, 2]` → `False`, so the field is *not* dropped and the round-trip silently breaks. The PR body documents the "natural Python form" expectation, but `StardagField.compat_default` itself is just `Any = _UNSET` with a `"""TODO"""` docstring. A one-line note on the field would prevent a future footgun. Not a blocker.

### Tests — good
Both reported repros covered, each with at-default (→ `{}`) and not-at-default (→ kept) cases, plus a no-mode baseline. Ran locally: 31 passed.

Minor: `enum.StrEnum` is 3.11+ while the package declares `requires-python = ">=3.10"`. Does not break CI (matrix is 3.11–3.14) and it is a test-only file, so it is fine — just flagging it would not run under a 3.10 interpreter.

### Verdict
**Approve.** Clean, minimal, correct fix with good regression coverage. Only optional follow-up: a one-line docstring note on `compat_default` about supplying values in validated Python form.

andhus:
Addressed all three points in a6bd60f:

1. **Short-circuit `hash_exclude`** (Copilot, base_model.py) — reordered to `hash_exclude or (...)` so `getattr(self, name)` is skipped when the field is dropped regardless.
2. **3.10-compatible enum** (Copilot + review, test_base_model.py) — replaced `enum.StrEnum` (3.11+) with `class Color(str, enum.Enum)` to match `requires-python = ">=3.10"`.
3. **Docstring on `compat_default`** (review) — documented `StardagField.compat_default`/`hash_exclude`, including the footgun: `compat_default` must be supplied in the field's natural validated Python form, since a non-idempotent input (e.g. `[1, 2]` for `tuple[int, ...]` → validated to `(1, 2)`) would fail the serialize-side equality check and silently not be dropped.

Tests still green (31 passed); ruff + pyright clean.

PR #150, opened 2026-06-12
Fix `stardag modal deploy` on modal >= 1.4.3 (`ensure_env` import)

### LINKED ISSUE #148: `stardag modal deploy` broken with modal>=1.4.3: cannot import `ensure_env` from `modal.environments`
## Summary

`stardag modal deploy <app.py>` fails at import time when the installed `modal` SDK is >= 1.4.3:

```
File ".../stardag/_cli/modal.py", line 541, in deploy
    from modal.environments import ensure_env
ImportError: cannot import name 'ensure_env' from 'modal.environments'
```

## Cause

modal 1.4.3 refactored `modal/environments.py` into a thin re-export wrapper around a new private `modal/_environments.py`, and `ensure_env` is no longer re-exported publicly. It now only exists as `modal._environments.ensure_env`.

In modal 1.4.2 the function was still defined directly in `modal/environments.py`, which is why this only surfaces after upgrading modal.

## Affected versions

- stardag 0.7.3 and 0.8.0 (checked the 0.8.0 wheel — `stardag/_cli/modal.py:541` has the same import)
- modal >= 1.4.3 (1.4.2 works)

## Reproduce

```
pip install "stardag[modal]==0.8.0" "modal==1.4.3"
stardag modal deploy some_app.py
```

## Workaround

Pin `modal==1.4.2`, or run the deploy with an ephemeral downgrade, e.g.:

```
uv run --with modal==1.4.2 stardag modal deploy some_app.py
```

## Possible fixes

- Import with a fallback: `try: from modal.environments import ensure_env / except ImportError: from modal._environments import ensure_env`
- Or inline the small env-name resolution logic to avoid depending on a private modal module going forward.

nivdee:
## Summary

`stardag modal deploy` crashed at import time on modal >= 1.4.3:

```
ImportError: cannot import name 'ensure_env' from 'modal.environments'
```

modal 1.4.3 refactored `modal/environments.py` into a thin re-export wrapper around the new private `modal/_environments.py`, and `ensure_env` is no longer re-exported publicly.

Fixes #148

## Approach

Instead of a `try/except` import fallback onto the private module, this inlines the small env-resolution logic using modal's **public** config API:

```python
if env is not None:
    modal_config.override_locally("environment", env)
env = modal_config.get("environment")
```

The body of modal's `ensure_env` is byte-identical across modal 1.0.0, 1.3.3, and 1.4.3 (verified against installed wheels), so this matches behavior across the full supported range (`modal>=1.0.0`) while removing any dependency on modal-internal modules going forward — the same class of breakage as #113.

## Testing

- New CLI regression test (`tests/test__cli/test_modal.py`) drives the `deploy` command past its modal imports and env resolution. Verified it **fails** with the old import on modal 1.4.3 (reproducing the exact `ImportError`) and **passes** with the fix on both modal 1.3.3 (lockfile) and 1.4.3.
- All other `modal.*` imports in the deploy path and integration module verified importable on 1.4.3.
- `tests/test__cli/` + non-network modal tests pass; pre-commit hooks (ruff, pyright, prettier) pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ client-code changes required.
   avoiding any dependency on modal-internal modules.
   ([#148](https://github.com/stardag-dev/stardag/issues/148),
   [#150](https://github.com/stardag-dev/stardag/pull/150))
+- Dev lockfile: bump modal 1.3.3 → 1.5.0 so CI continuously tests against a
+  modal version affected by the import breakage. The supported range is
+  unchanged (`modal>=1.0.0`).
 
 ## [0.8.0] — 2026-06-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,15 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
 ## [Unreleased]
 
+### SDK
+
+- **`stardag/_cli/modal.py`**: Fix `stardag modal deploy` crashing with
+  `ImportError: cannot import name 'ensure_env' from 'modal.environments'` on
+  modal >= 1.4.3, where `ensure_env` moved to a private module. The small
+  environment-resolution logic is now inlined using modal's public config API,
+  avoiding any dependency on modal-internal modules.
+  ([#148](https://github.com/stardag-dev/stardag/issues/148))
+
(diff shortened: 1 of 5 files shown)


andhus (approved):
Let's prepare a v0.8.1 release 👍

PR #171, opened 2026-07-13
Move reactive build metadata to the registry

andhus:
## What

Move reactive Modal scheduling's per-build orchestration metadata — the reactive **marker**, owning **app_name**, and **tick_kwargs** — off the default target root (a `meta.json` written by `BuildTaskStore.write_meta`/`read_meta`) and into the **registry**.

Task pickles (user task *objects*) stay on the target root — the server must never materialize user task objects. This PR moves only the marker + app_name + tick_kwargs.

## Why

Target roots are the **asset store** and may be configured immutable/append-only (S3 object-lock; Modal volumes refuse overwrites), so rewriting `meta.json` on re-trigger crashed. It was also split-brain: the registry is already the source of truth for build roots/status/frontier.

## Changes

**Registry API (stardag-api)**

- New nullable `builds.reactive_meta` JSONB column — `{"app_name": str, "tick_kwargs": {...}}`; `NULL` = not reactively scheduled (presence is the marker). One additive/nullable migration (`20260714_011326_add_build_reactive_meta.py`, verified upgrade→downgrade→upgrade on Postgres).
- `PUT /api/v1/builds/{build_id}/reactive-meta` — upsert (idempotent; a re-trigger may update `tick_kwargs`), env-scoped auth, rate-limited, size-capped, consistent with neighboring build mutations.
- `reactive_meta` exposed on `BuildResponse` and on `BuildFrontierResponse` (so a tick gets it in the frontier call it already makes).

**SDK (lib/stardag)**

- `RegistryABC` + `APIRegistry`: `build_set_reactive_meta[_aio](build_id, *, app_name, tick_kwargs)`; `NoOpRegistry` no-ops. `reactive_meta` added to the `BuildFrontier` model.
- The reactive trigger sets `reactive_meta` via the registry instead of writing `meta.json`. Because the registry is mutable, **a re-trigger may now update `tick_kwargs`** (previously fixed at first trigger).
- The tick reads the marker / owner / config from the frontier's `reactive_meta`; the ownership guard compares `reactive_meta["app_name"]`.
- `BuildTaskStore` is now **pickle-only** (`write_meta`/`read_meta` removed; module docstring updated).
- **Old server:** reactive scheduling already required a matching stardag-api version; `reactive_meta` folds into that same contract — an older server fails the reactive trigger clearly (like the existing "does not support reactive scheduling" path), no target-root fallback.

**Docs / changelog**

- `CHANGELOG.md` (SDK + Registry API sections), the Modal how-to reactive section, and the build-execution concepts page.

## Tests

- API: set/get `reactive_meta`, it appears on the build response + frontier, non-reactive build is null, build-not-found 404, environment isolation (403). Migration up/down verified on Postgres.
- SDK: reactive trigger sets it via the registry; tick reads marker/ownership from the frontier `reactive_meta` (fakes extended in `test_reactive.py` and `test_stardag_app.py`, migrated off the old `write_meta` seeding); re-trigger updates `tick_kwargs` (previously impossible).
- Both suites green: `lib/stardag` 794 passed; `app/stardag-api` 304 passed. pyright clean on both packages.

## Merge notes — DO NOT MERGE yet

- **Stacked** on the open SDK-fixes PR (base `fix/resource-provider-cloudpickle`). **Rebase onto `main` after that PR merges.**
- Requires a **server deploy** (new column + endpoint) before the SDK change is used against production.


andhus:
Rebased and reconciled onto `main`.

Previously this branch was stacked on the pre-squash resource-provider/write-once branch. `main` has since absorbed the write-once build store and the `stardag_api_key_secret`/Modal-workspace-bake work as squashed commits, so I replayed only this PR's unique change onto current `main` (`--onto`). Net diff is now exactly the reactive-meta move; base retargeted to `main`.

Reconciliation notes:
- `BuildTaskStore`: kept `main`'s write-once `save_task` (skip-if-exists); removed `write_meta`/`read_meta` (marker now lives in the registry). Store is pickle-only.
- `integration/modal/_app.py`: kept the `finalize()` api-key-secret propagation and `STARDAG_MODAL_WORKSPACE` bake untouched; the tick now reads the reactive marker/owner/config from the frontier's `reactive_meta` (registry) instead of the store meta, and the trigger upserts it via `registry.build_set_reactive_meta`.
- `_reactive.py`: reactive marker check reads `frontier.reactive_meta`.
- CHANGELOG: relocated this PR's entry from the already-released 0.10.1 section into a new **0.11.0** section (0.10.1/0.10.2 are tagged).

Carries the additive, nullable `builds.reactive_meta` JSONB migration (re-parented onto `main`'s current head; upgrade/downgrade/upgrade verified on Postgres). **Requires deploying the registry with this migration applied** before an SDK new enough to PUT the reactive-meta endpoint drives reactive builds.

SDK tests: 625 passed / 19 skipped (`-m 'not modal_live'`). API tests: 304 passed / 18 skipped. Lint (ruff, ruff-format, prettier, pyright) green.

andhus:
Review in two parts, per request: first the design question (JSONB vs structured), then correctness of the change as written.

## Design: is JSONB the right shape for this metadata?

The schema already follows an implicit promotion rule, and it's the right one — worth stating explicitly somewhere (concepts page or a models.py comment):

> **Fields the server must enforce, join, or filter on get typed columns/tables; fields it merely records and echoes stay JSONB.**

That's exactly why `latest_executor`/`latest_executor_ref` are typed columns (re-attach logic needs them), `task_limit_keys`/`environment_concurrency_limits` are real tables (slot counting is a server-side transactional concern), while `event_metadata`, `executor_metadata`, and `task_data` are JSONB (server-opaque, SDK-owned shapes).

Applying it to the three things this PR stores:

- **`tick_kwargs` — JSONB is right.** It's SDK-owned free-form config that evolves with `TickConfig`; typed columns (or server validation beyond the size cap) would couple server releases to SDK config evolution. Keep opaque + capped.
- **The marker — NULL-presence is right.** `reactive_meta IS NOT NULL` is filterable without JSONB path ops. But that only pays off if the list endpoint actually gains the filter — it didn't in this PR (see findings): the watchdog still lists all builds and now pays a frontier call per build to discover non-reactive ones. The registry-side marker was motivated precisely by that query.
- **`app_name` — the one field on the wrong side of the line.** Today only the SDK's ownership guard reads it, so JSONB works. But the natural next query is server-side: "list running reactive builds owned by app X" (the watchdog's actual question, and the fix for its full-history pagination). That filters on `app_name` → typed-column territory. Since the migration is being cut *now*, promoting it is free: `builds.reactive_app_name TEXT NULL` (presence = marker, owner, filterable, indexable) + `builds.tick_kwargs JSONB NULL` — same single migration, no second migration later, and the PUT/response shapes can stay as they are.

Alternatives considered and rightly not taken: a 1:1 `build_scheduling` table (right shape *if* server-side scheduling state grows — lease bookkeeping, watchdog timestamps — but today it just adds a join to every frontier/list read); per-executor structured tables (contradicts the executor-agnostic server — every new executor backend would need server migrations); fully-typed columns including tick knobs (every new knob = a migration).

**Verdict: keep JSONB for `tick_kwargs`; consider promoting `app_name` to a typed nullable column in this same migration; add the `reactive` list filter now or soon.**

## Correctness findings

**major — in-flight reactive builds from ≤0.10.x are orphaned by the upgrade; no fallback, no docs, no test.** The tick gate is now exclusively `frontier.reactive_meta is None → not_reactive` and `BuildTaskStore.read_meta` is deleted. A build triggered by the previous SDK has `meta.json` but no registry row: after the same-name app redeploy (the supported upgrade path), every worker wake-up and watchdog sweep short-circuits `not_reactive` — the build silently stops being scheduled, with `needs_tick_at` set and no error anywhere. Same stall for the skew "new-SDK client triggers against a not-yet-redeployed old app". The remedy (re-trigger — PUTs the meta; pickles are write-once so re-persist is safe) works but is documented nowhere. Minimum: an upgrade note in CHANGELOG/how-to ("re-trigger in-flight reactive builds after upgrading"). Nicer: a transitional fallback in the tick wrapper — if `reactive_meta is None`, read the legacy `meta.json` once and PUT it (auto-migrates in one tick).

**major — bare re-trigger resets `tick_kwargs`, regressing the 0.10.1 merge-semantics fix.** (verified) The trigger sends `tick_kwargs=tick_kwargs or {}` unconditionally and the server PUT full-replaces the dict. A build triggered with `tick_kwargs={"fail_mode": "continue"}` and later re-triggered bare — the documented takeover recipe and the stalled-build wake — silently flips back to default FAIL_FAST. This is exactly the regression the earlier "meta is MERGED" fix closed; the test that pinned it was rewritten to re-trigger *with* explicit kwargs, so the bare case is now untested, and the `test_stardag_app.py` class docstring still says "merge store meta". Fix: omit `tick_kwargs` from the PUT when not explicitly passed and have the server preserve the stored value when the field is absent (omit-means-preserve), plus restore the bare-re-trigger test.

**minor — the pre-lease gate costs a full frontier query per tick invocation, and per running build per watchdog sweep per app — including resident and foreign builds.** The tick wrapper fetches the frontier for the marker/ownership check, then `run_tick_aio` clears-and-refetches inside the lease — 2+ frontier calls (the documented-expensive endpoint) per legitimate wake-up, and the watchdog now pays one per *resident* build per sweep where the old gate was a target-root read costing the server nothing. `reactive_meta` is already on `BuildResponse` — use `GET /builds/{id}` for the entry gate, or thread the entry frontier into the first loop iteration.

**minor — `build_list_running` gained no `reactive` filter** (see design section) — the one-line `reactive_meta IS NOT NULL` WHERE clause would let the watchdog skip non-reactive builds server-side instead of discovering them one frontier call at a time.

**minor — `RegistryABC.build_set_reactive_meta` defaults to a silent no-op.** A custom registry backend that supports the frontier but predates the method makes the reactive trigger "succeed" while every tick sees `reactive_meta is None` and no-ops forever — contradicting this PR's own "fails clearly rather than degrading silently" contract (which holds only for `APIRegistry` via the 404 mapping). Raise `NotImplementedError` by default, or read-back-verify at the trigger.

**nits:** no way to clear/unset `reactive_meta` once set (no DELETE, schema can't express null) — same one-way property as the old `meta.json`, harmless today, worth a docstring sentence; `tick_kwargs` is persisted free-form and a malformed dict makes `_build_tick_config` raise on every subsequent tick of that build (pre-existing risk, but the server owning the data makes a shape check possible now); the PUT mutates the build row with no event recorded — consistent with the resume-with-metadata precedent, deserves the same code comment; **no CI checks reported on this branch** after the rebase — re-push or re-run before merge.

**test gaps:** the in-flight upgrade scenario; bare-re-trigger-preserves-tick_kwargs (regressed and untested).

**Verified sound:** migration is a single nullable JSONB add off the current main head, single-headed, correct downgrade, instant on live Postgres; PUT is env-scoped (404/403 isolation tested, value untouched on 403), rate-limited, size-capped (4 KB compact JSON bounding app_name + tick_kwargs together), idempotent last-write-wins; all 10 `BuildResponse` construction sites gained the field; frontier exposure costs zero extra queries; `build_set_reactive_meta` is called before the first tick spawn (no trigger/tick race); the gate's frontier GET does not clear `needs_tick_at`, so foreign/stray ticks lose no wake-ups; the forward-to-owner path is intact incl. owner-app-gone; `save_task` is write-once, so re-trigger re-persistence genuinely works on immutable/append-only roots — the motivating bug is fixed for pickles too; old SDK + new server unaffected; new SDK + old server fails loudly at both trigger and tick (route-404 vs resource-404 correctly discriminated).

**Deploy/merge verdict:** the move is architecturally right (registry as source of truth for orchestration state; target root as immutable asset store) and the server side is deploy-safe any time. Not upgrade-safe as written, though: the two majors — silent orphaning of in-flight builds and the tick_kwargs reset on the documented recovery gesture — should land before merge; both fixes are small. The release-note line "server deploy required before SDK use" is correct and should gain the in-flight re-trigger instruction.

andhus:
## Review feedback addressed (pushed `3a50620`)

Reworked the still-unreleased migration and reactive-metadata design per review. All CI checks green; SDK `801 passed / 6 skipped`, stardag-api `306 passed / 18 skipped`. Migration verified up→down→up on a throwaway Postgres.

### A — Promote `app_name` to a typed column + list filter ✅
- Split the JSONB `reactive_meta` into two columns: **`builds.reactive_app_name`** (indexed `String(64)` — owner **and** marker; `reactive_app_name IS NOT NULL` is now the "reactively scheduled" test) and **`builds.reactive_tick_kwargs`** (JSONB, SDK-owned `TickConfig` kwargs). All marker checks updated (SDK + server).
- Reworked the unreleased migration to add both columns + the index (additive/nullable, instant). Verified base→head, head→down (columns/index gone), down→head again on Postgres 16.
- `GET /api/v1/builds` gains **`reactive_app_name`** (SQL filter on the indexed column, correct pagination/count) and **`status`** filters, e.g. `?reactive_app_name=<app>&status=running`. `status` is derived from events (not a column) so it can't be paginated in SQL — matching candidates are scanned up to a bounded cap (500, logged on truncation) and filtered in Python; intended to be combined with `reactive_app_name`, which keeps the set tiny. Both fields exposed on the build + frontier responses. New API test covers the app+status combination.
- **A.4 watchdog rewiring — deferred (deliberate).** The filter endpoint exists and is tested, and a client `build_get` is wired. I did **not** repoint the watchdog at `reactive_app_name=<self>&status=running`: today every app's watchdog sweeps *all* running builds so a stray wake-up landing on the wrong app (e.g. a previous owner's still-running worker after a takeover) is still forwarded — scoping the sweep to the app's own builds removes that cross-app coverage. Follow-up: revisit whether owner-only sweeping (each owner covers its own builds) is sufficient, and if so switch the watchdog to the new filter.

### B — In-flight builds orphaned on upgrade ✅ (release note only)
- Added a prominent **⚠️ upgrade warning** to `RELEASE_NOTES.md` (and CHANGELOG): reactive builds in flight across the upgrade have no registry marker and their ticks silently no-op, so they must be **re-triggered** (`build_trigger(..., build_id=<id>, reactive=True)`).
- **Self-heal intentionally skipped** as low value: it would require re-adding a legacy `meta.json` read path, and the clean registry-only cut is the desired end state. (No pre-`0.11.0` in-flight builds are actually at stake.)

### C — Bare re-trigger no longer wipes `tick_kwargs` ✅
- `SetReactiveMetaRequest.tick_kwargs` is now `dict | None = None`. The server sets `reactive_app_name` always but only writes `reactive_tick_kwargs` when `tick_kwargs` is provided — a bare re-trigger **preserves** the stored config; a re-trigger *with* `tick_kwargs` updates it.
- SDK side: the registry client omits `tick_kwargs` from the PUT body when `None`, and the reactive trigger passes it straight through (no more `tick_kwargs or {}`). Regression tests pin both the server behavior (`test_bare_retrigger_preserves_tick_kwargs`) and the SDK call (bare re-trigger → `tick_kwargs=None`).

### D — Pre-lease gate no longer does a full frontier query ✅
- Added `build_get` / `build_get_aio` (→ `GET /builds/{id}`, returning a slim `BuildInfo` with the two reactive fields). The Modal tick's pre-lease marker/ownership gate now uses `build_get`; the full `build_get_frontier` is fetched only inside `run_tick_aio` where the frontier is actually processed (the backstop marker re-check there reuses that same fetch — no extra query).

### E — Copilot inline comments ✅
1. `schemas.py` — `tick_kwargs` mutable `{}` default removed (now `None`, which also gives the C merge semantics).
2. `_app.py` — the `build_get_frontier`→`build_get` gate now only rewrites a **genuine missing-route 404** into "server too old"; a resource-level 404 (build deleted) re-raises. Uses a shared `is_missing_route_error` helper.
3. `_reactive.py` — same fix on the frontier/notify 404 path (only `detail == "Not Found"` is converted; real 404s propagate). The over-broad rewrite is gone from both paths.

### F — CI ✅
CI now triggers on the branch (it hadn't after the earlier force-push). Latest run is green across all 9 checks (lint, Python 3.11–3.14, stardag-api on Postgres, frontend, integration + browser).

Issue #173, opened 2026-07-14
UI: Modal function-call deep link uses the wrong dashboard URL format


andhus:
## Problem

The task/build **executor badge** deep links to a Modal function call via `src/utils/modalLinks.ts` (`modalFunctionCallUrl`), which builds:

```
https://modal.com/apps/<workspace>/<env>/deployed/<app>?functionCallId=<fc-id>
```

That query param is not what the Modal dashboard uses, so the link doesn't land on the function call. The **app-level** link (`.../deployed/<app>`) is correct.

## Correct format

A working function-call URL looks like:

```
.../deployed/<app>?activeTab=functions&functionId=<fu-id>&functionSection=calls&fcId=<fc-id>
```

(the `&start=…&end=…&live=true` and `&inputId=…` params can be stripped and it still resolves).

## Scoping — one open question

The correct URL needs the **function id** (`fu-…`), i.e. the deployed worker function's id — the SDK currently records only the function *name* and the function *call* id (`fc-…`) in executor metadata.

- **If the URL resolves without `functionId`** (i.e. `?activeTab=functions&functionSection=calls&fcId=<fc-id>` works) → **UI-only fix**: rewrite `modalFunctionCallUrl`. No SDK/metadata change.
- **If `functionId` is required** → also capture the worker `modal.Function`'s `object_id` (`fu-…`; available via `Function.hydrate().object_id`) into the executor metadata (SDK), expose it in the API/UI, and build the URL from it.

## Fix location & deploy

`modalFunctionCallUrl` lives in the UI (`app/stardag-ui`), so this needs a UI change + production UI deploy; the SDK-metadata half (if needed) rides a future SDK release. Keep all Modal URL patterns centralized in `modalLinks.ts`.

Discovered while validating the reactive Modal walkthrough end-to-end (0.10.2).

andhus:
## Link stability across app stop/redeploy

The `.../deployed/<app-name>` path only resolves while that app version is **live**. Once an app is stopped or redeployed, the dashboard addresses it by **app id** instead:

```
https://modal.com/apps/<workspace>/<env>/ap-<app-id>?activeTab=functions&functionId=fu-…&functionSection=calls&fcId=fc-…&inputId=in-…
```

So for links that survive a stop/redeploy we should prefer the **app-id form** and capture as many identifiers as we can at execution time:

- **app id** (`ap-…`) — from the deployed `modal.App` / the function's app.
[...]

PR #175, opened 2026-07-14
Capture Modal app_id + function_id in executor metadata (stable deep links)

andhus:
## Summary

The Modal executor metadata dict (previously `kind, app_name, workspace, environment, function_name`) now also captures two Modal object identifiers so the UI can build stable, stop/redeploy-proof dashboard deep links:

- **`app_id`** (`ap-…`) — app-wide, added to the *base* executor metadata. Resolved once and cached via best-effort `modal.App.lookup(app_name, environment_name=…).app_id` (works both locally and in-container).
- **`function_id`** (`fu-…`) — per-worker, added next to `function_name`. Resolved by hydrating the worker `modal.Function` handle and reading `.object_id`.

## Worker propagation

Workers self-report their lifecycle with this metadata. Both new fields ride the existing executor-metadata env-override channel: new `STARDAG_MODAL_APP_ID` / `STARDAG_MODAL_FUNCTION_ID` env constants are wired in both directions — the executor→worker override map and the worker-side `_WorkerLifecycleReporter` read map — so a worker's self-reported `TASK_STARTED` carries `app_id`/`function_id` too.

## Best-effort

Resolution never fails or delays a task start. On any error (broken token, unreachable Modal API, hydration failure) the offending key is simply omitted and the failure logged at debug — matching the existing pattern used for `workspace`.

## Why

This feeds the UI dashboard deep-link fix (issue #173). The app-id URL form (`.../apps/{workspace}/{env}/{app_id}?...&functionId={function_id}&fcId={fc}`) keeps resolving after an app version is stopped or redeployed, unlike the `deployed/{app-name}` form. The `fc-…` function-call id was already recorded as the executor ref; this PR adds the remaining two ids.

## Tests

Hermetic unit tier extended: the `hermetic_modal_executor_metadata` conftest fixture now also pins `app_id`/`function_id` (so no test hits the network), and a new test class exercises the real resolvers with a monkeypatched `modal.App.lookup` / handle hydration — covering both keys present when resolvable, omitted on failure, and propagated to the worker env + read back by the reporter. Full non-live suite green (`pytest tests -m "not modal_live"`: 798 passed, 2 skipped).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,13 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
   self-reported start carries them too. They let the UI build stable
   dashboard deep links in the app-id URL form, which keeps resolving after
   an app version is stopped or redeployed (the deployed-app-name form does
-  not). Best-effort throughout: resolution never fails or delays a task
-  start — on any error the key is simply omitted.
+  not). Best-effort throughout: on any error the key is simply omitted rather
+  than raised, so resolution never fails a task start. The two lookups sit on
+  the critical path before `spawn`, so they can add latency — but each is
+  bounded by a short (3 s) timeout, so a slow or hung Modal API cannot stall
+  a start beyond that cap. Resolved values (including a resolved-but-missing
+  id) are cached per process, so a failing lookup is not re-paid on every
+  start.
 
 ## [0.10.2] — 2026-07-14
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,20 @@ All notable changes to the Stardag project (SDK, Registry API, and UI).
 
 For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
+## [Unreleased]
+
+### SDK
+
+- **Modal executor metadata now records the app id (`app_id`, `ap-…`) and
+  worker function id (`function_id`, `fu-…`).** These ride the existing
+  executor-metadata channel (base metadata + the worker env-override
+  propagation, read back by the worker lifecycle reporter) so a worker's
+  self-reported start carries them too. They let the UI build stable
+  dashboard deep links in the app-id URL form, which keeps resolving after
+  an app version is stopped or redeployed (the deployed-app-name form does
+  not). Best-effort throughout: resolution never fails or delays a task
+  start — on any error the key is simply omitted.
+
 ## [0.10.2] — 2026-07-14
 
 ### SDK
--- a/lib/stardag/src/stardag/integration/modal/_app.py
+++ b/lib/stardag/src/stardag/integration/modal/_app.py
@@ -628,6 +628,14 @@ def _get_modal_environment() -> str | None:
     return modal_config.get("environment") or None
 
 
+# Upper bound (seconds) on the best-effort Modal id lookups performed on the
+# critical path before ``spawn``. Matches the 3 s deadline Modal's own
+# ``_lookup_workspace`` gRPC call uses: a *hung* (not merely refused) Modal
+# API must not stall the first task start beyond this cap. On timeout the id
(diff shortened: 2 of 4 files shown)


andhus:
Reviewed together with its UI counterpart (fix/modal-deeplinks-and-nav). Correct and consistent with the established metadata patterns — env-override wiring is symmetric both directions with no warm-container leakage, `create_if_missing` stays False, the hermetic fixture pins both new resolvers (no network in the unit tier), key names match the UI's `ExecutorMetadata` exactly, and worker propagation + reporter read-back are tested end-to-end. Nothing blocking; two hardening items on the best-effort resolvers:

**minor — `App.lookup` / `Function.hydrate` are unbounded, unlike the workspace-lookup precedent.** Neither takes a timeout (verified against the signatures) — the gRPC call runs under Modal's default retry/deadline policy, so on a *hung* (not refused) Modal API the first task start can stall well past the 3 s bound this module deliberately imposed on `_lookup_workspace`; the "never delays a task start" claim doesn't hold for that first start. Shared fate with the spawn softens it, but wrap both in `asyncio.wait_for(..., 3)` for symmetry with the documented pattern.

**minor — function-id hydration failure is not cached.** App-id failure is effectively cached via the memoized base dict, but a persistently failing `hydrate` re-pays a network attempt on every task start — the exact per-start re-pay pattern the earlier workspace-lookup review fix eliminated. Cache per worker name, success and failure.

**nit — wrong-app `app_id` on unset environment.** `App.lookup(environment_name=None)` resolves the *config-default* environment: an app deployed to env X, with local config defaulting to env Y that also has a same-name app, records Y's `app_id` — making the UI's "stable" deep link stably wrong. Narrow (needs same-name apps in two envs + unset `environment`), and the `environment` key is already equally wrong in that scenario — a docstring caveat suffices.

andhus:
Addressed the review feedback in 889989c (best-effort resolvers hardened; no behavior change to the happy path).

**1 — Bound both id lookups with a short timeout (human + Copilot).** Both `modal.App.lookup` (app id) and `Function.hydrate` (function id) now run under `asyncio.wait_for(..., timeout=_MODAL_ID_LOOKUP_TIMEOUT_SECONDS)`, with the constant set to `3.0` — matching the deadline Modal's own workspace lookup gRPC call uses. A timeout is caught like any other failure: the key is omitted and the miss logged at debug, never raised. So a *hung* (not merely refused) Modal API can no longer stall the first task start past the 3 s cap.

**2 — Cache failures, not just successes (human).** Function-id resolution is now cached per worker name for **both** outcomes via a new `_worker_function_ids: dict[str, str | None]` on the executor — membership (not truthiness) marks "resolved", so a resolved-but-`None` (failed/timed-out hydration) is kept and not re-attempted on the next start. App-id was already effectively cached via the memoized base-metadata dict (a missing `app_id` key persists), so no change was needed there. Cache keys respect scope: app id is app-wide (base dict), function id is per-worker.

**3 — CHANGELOG wording (Copilot).** Reworded to drop the over-broad "never delays a task start". It now states the two lookups sit on the critical path before `spawn` and *can* add latency, but each is bounded by a short (3 s) timeout so a slow/hung Modal API cannot stall a start beyond that cap; on any error the key is simply omitted (best-effort, never raised), and resolved values (including a resolved-but-missing id) are cached per process.

**nit — wrong-app `app_id` on unset environment (human).** Added a docstring caveat to `_get_modal_app_id_aio` documenting that `environment_name=None` resolves the config-default environment (the `environment` metadata key is equally affected in that same-name-app-in-two-envs scenario); pass the resolved environment to avoid it.

**Tests:** added timeout-path coverage for both resolvers (patched constant + a hanging lookup/hydrate → `None`) and function-id caching tests (positive: resolver called once across two submits; negative: a failed resolution is cached, not retried, and the key stays omitted). Full non-live suite green: `pytest tests -m "not modal_live"` → 808 passed, 6 skipped.

Issue #174, opened 2026-07-14
UI: 'Concurrency Limits' sidebar nav item wraps and is center-aligned


andhus:
## Problem

The **Concurrency Limits** item in the left sidebar navigation wraps to two lines, and the wrapped text renders **center-aligned** — unlike every other nav item, which is single-line and left-aligned. Looks broken next to the others.

## Fix

Two parts (either alone helps; both is best):

1. **Shorten the label** so it fits one line — e.g. `Concurrency` or `Conc. Limits`.
2. **Make the nav label robustly left-aligned** (and ideally `white-space: nowrap` / truncate with ellipsis) so a long label stays consistent with the others instead of centering.

## Location

`app/stardag-ui` — the sidebar `NavItem` for the concurrency-limits route (added with the concurrency-limits admin view). UI-only; needs a production UI deploy. Can be batched with the deep-link fix (#173) in one UI deploy.

PR #176, opened 2026-07-14
UI: stable Modal deep links + execution details + sidebar nav label

andhus:
Fixes two UI issues in `app/stardag-ui`. UI-only; needs a production UI deploy.

## 1. Stable Modal function-call deep links (#173)

The task-detail and concurrency-holder "View on Modal" links used a query-param form (`?functionCallId=…`) that doesn't resolve in the Modal dashboard, and even the working `deployed/{app_name}` form only resolves while that app version is live — a stopped/redeployed app is addressed by its **app id** instead.

`modalFunctionCallUrl` now picks the most resilient form it has enough identifiers for:

1. `workspace + app_id + function_id + fcId` → **stable, stop/redeploy-proof** app-id URL:
   `https://modal.com/apps/{workspace}/{environment}/{app_id}?activeTab=functions&functionId={function_id}&functionSection=calls&fcId={fc}`
2. else `workspace + app_name + function_id + fcId` → the `deployed/{app_name}?activeTab=functions&functionId=…&functionSection=calls&fcId=…` form (works while the app is live).
3. else → the plain app-page link (or `null` if even that can't be built).

Every segment/param is `encodeURIComponent`-escaped, and all Modal URL patterns stay centralized in `modalLinks.ts` (Modal gives no URL-format guarantee). New optional `app_id` (`ap-…`) / `function_id` (`fu-…`) keys were added to the `ExecutorMetadata` type and are read **defensively** — absent on older data, which still degrades to a working, non-dead link.

> Pairs with the SDK metadata PR (`feat/modal-id-metadata`), which starts recording `app_id`/`function_id` in the Modal executor metadata. This UI reads them if present and works fine without them.

## 2. "More details" block in TaskDetail (#173)

The Execution section gains a collapsible **"more details"** block listing every captured Modal identifier verbatim (kind, app/function names, workspace, environment, app id, function id, and the function-call ref), each click-to-copy — so a user can reconstruct a reference by hand even if the dashboard URL format drifts. Only present fields render.

## 3. Sidebar nav label (#174)

Shortened "Concurrency Limits" to **"Concurrency"** and made every nav label robustly left-aligned and single-line (`truncate` with ellipsis) so a long label never wraps and center-aligns like the others.

## Tests / quality

- `modalLinks`: stable URL with all ids, deployed-name fallback, plain-app fallback, null-safety, non-modal kind, URL-encoding.
- `ModalExecutionDetails`: present/absent fields, empty render, click-to-copy.
- `Sidebar`: label rendering + left-align/no-wrap classes.
- Full UI suite green (57 tests). prettier / eslint / tsc clean.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,29 @@ All notable changes to the Stardag project (SDK, Registry API, and UI).
 
 For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
+## [Unreleased]
+
+### UI
+
+- **Stable, stop/redeploy-proof Modal function-call deep links.** The task
+  detail and concurrency-holder "View on Modal" links previously used a
+  query-param form that didn't resolve. They now build the app-id URL
+  (`.../apps/{workspace}/{env}/{app_id}?activeTab=functions&functionId=…&functionSection=calls&fcId=…`)
+  when the newly captured `app_id`/`function_id` metadata is present,
+  falling back to the deployed-app-name form and then the plain app-page
+  link. Reads the new metadata defensively, so older data without the ids
+  still gets a working, non-dead link. Pairs with the SDK change that
+  records `app_id`/`function_id` in the executor metadata.
+- **Task detail: "more details" block for Modal identifiers.** The
+  Execution section now has a collapsible list of every captured Modal
+  identifier (kind, app/function names, workspace, environment, app id,
+  function id, and the function-call ref), each click-to-copy, so a
+  reference can be reconstructed by hand if the dashboard URL format
+  drifts. Only present fields render.
+- **Sidebar: shortened the "Concurrency Limits" nav item to "Concurrency"
+  and made every nav label left-aligned and single-line (truncating with
+  an ellipsis instead of wrapping and centering).**
+
 ## [0.10.2] — 2026-07-14
 
 ### SDK
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,18 +14,26 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
   (`.../apps/{workspace}/{env}/{app_id}?activeTab=functions&functionId=…&functionSection=calls&fcId=…`)
   when the newly captured `app_id`/`function_id` metadata is present,
   falling back to the deployed-app-name form and then the plain app-page
-  link. Reads the new metadata defensively, so older data without the ids
-  still gets a working, non-dead link. Pairs with the SDK change that
-  records `app_id`/`function_id` in the executor metadata.
+  link. The app-page fallback itself prefers the stable app-id form
+  (`.../apps/{workspace}/{env}/{app_id}`) whenever `app_id` is available —
+  so metadata with an `app_id` but no `function_id` still degrades to a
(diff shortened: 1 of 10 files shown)


andhus:
Reviewed together with its SDK counterpart (feat/modal-id-metadata). Clean fix — the fallback ladder is correct and defensively coded, every segment/param is `encodeURIComponent`-escaped, the broken `?functionCallId=` form is fully gone (both consumers — TaskDetail and the concurrency-holders view — route through `modalFunctionCallUrl`, and the module comment forbids reintroducing it), old data without the new ids still renders working links, no `dangerouslySetInnerHTML` anywhere, and coverage matches the PR body (16 modalLinks tests incl. the full ladder and URL-encoding). Safe to deploy before or after the SDK release. Two nits:

**nit — `app_id`-without-`function_id` degrades past the stable form.** Both deep forms require `function_id`, so metadata with `app_id` but no `function_id` falls back to the deployed-name app page — exactly the form that dies on stop/redeploy — while a stable app-page URL (`…/{app_id}` without the function params) may have been buildable. If the app-id URL resolves without the query params (one manual check), add that rung to the ladder.

**nit — a truncated expanded-sidebar label has no tooltip.** `title` is set only when `collapsed`; with the new `truncate`, a label that ellipsizes in the expanded sidebar can't be read by hover. Trivial: `title={item.label}` unconditionally.

andhus:
Addressed all three review items in ba69048 (pushed to `fix/modal-deeplinks-and-nav`; not merged).

**1. Gate the "more details" block by executor kind** (Copilot, `TaskDetail.tsx`)
`ModalExecutionDetails` now returns null for an explicitly non-modal `kind` (e.g. `k8s`), so its Modal-specific labels ("App name", "Function ID", …) never mislabel unrelated identifiers. It still renders for modal metadata, the legacy kind-less case (treated as modal for back-compat), and the executorRef-only case (no metadata). Reuses the exported `isModalMetadata` guard. New tests cover the non-modal-kind and kind-less cases.

**2. Missing stable app-page fallback rung** (`modalLinks.ts`)
`modalAppUrl` now prefers the stable app-id form (`.../apps/{workspace}/{env}/{app_id}`) whenever `app_id` is present, degrading to the deployed-name form and then null. This closes the nit: metadata with an `app_id` but no `function_id` now degrades to a stop/redeploy-proof app page rather than the deployed-name page that dies on stop/redeploy — and `modalFunctionCallUrl` inherits it since its rung 3 calls `modalAppUrl`. Graceful degradation for older data (no app_id → deployed-name → null) is preserved. Tests updated + added rungs for both fallbacks.

**3. Tooltip on truncated nav labels** (`Sidebar.tsx`)
`title={item.label}` is now set unconditionally (was gated on `collapsed`), so a label that ellipsizes under `truncate` in the expanded sidebar stays readable on hover. Added a `Sidebar.test.tsx` assertion that every item carries `title={label}`.

Quality: prettier / eslint / tsc clean; full UI suite green (64 tests, +7 vs before). CHANGELOG updated.

andhus:
Follow-up (8ec144d): the item-3 tooltip surfaced a Playwright strict-mode collision in the browser integration tests — the nav label now carries both its text and a matching `title`, so the old `get_by_text(label).or_(get_by_title(label))` locator matched two elements. Switched those locators to `get_by_role("button", name=...)`, which resolves to a single element whether the sidebar is expanded (name from label) or collapsed (name from title). All CI green now.

PR #177, opened 2026-07-14
CLI: concurrency-limit CRUD (list/set/delete/holders/evict)

andhus:
Adds a `stardag concurrency-limits` command group so users can manage named
concurrency limits from the CLI instead of an ad-hoc script.

## Commands

All accept `-p/--stardag-profile` and `-e/--stardag-env` to target a
non-active profile / environment.

- `stardag concurrency-limits list [--holders]` — table of key /
  max_concurrent; `--holders` adds each key's current holder count (one extra
  call per key).
- `stardag concurrency-limits set <key> <max_concurrent>` — upsert
  (`PUT /api/v1/concurrency-limits/{key}`); `max_concurrent` must be >= 1.
- `stardag concurrency-limits delete <key> [--yes]` — DELETE (confirm unless
  `--yes`).
- `stardag concurrency-limits holders <key> [--limit N]` — list RUNNING slot
  holders (task id/name, running-since, executor), oldest-running first.
- `stardag concurrency-limits evict <key> <task_id> [--yes]` — POST evict to
  free a leaked slot (confirm unless `--yes`).

## Implementation

- New `APIRegistry.concurrency_limit_{list,set,delete,holders,evict}` methods
  wrap the endpoints and reuse the client's rate-limit retry + error handling
  (preferred over calling httpx directly for testability).
- New `_cli/limits.py` command group, registered as `concurrency-limits` in
  the CLI; resolves the active/overridden profile + environment and builds an
  `APIRegistry` (with the same "no registry configured" guard the ad-hoc
  script used).
- Tests: typer `CliRunner` against a mocked registry — list/set/delete/holders/
  evict happy paths, confirmation-abort, validation, API-error reporting, and
  the no-registry guard.
- Docs: a Concurrency Limits section in the CLI reference, an entry in the SDK
  skill bundle, and a CHANGELOG `[Unreleased]` note.

This replaces the walkthrough example's ad-hoc `configure_limits.py`, which
can be retired once this ships (its migration to the CLI is left as a separate
follow-up and is not touched here).

### DIFF
--- a/.claude/skills/stardag/registry-and-platform.md
+++ b/.claude/skills/stardag/registry-and-platform.md
@@ -145,6 +145,19 @@ stardag environment target-roots add KEY URI # Add target root
 stardag environment target-roots list        # List target roots
 ```
 
+### Concurrency Limits
+
+```bash
+stardag concurrency-limits list              # List named limits (--holders adds counts)
+stardag concurrency-limits set KEY N         # Upsert a limit (max_concurrent = N)
+stardag concurrency-limits delete KEY        # Remove a limit (--yes to skip confirm)
+stardag concurrency-limits holders KEY       # List RUNNING slot holders
+stardag concurrency-limits evict KEY TASK_ID # Free a leaked slot (--yes to skip confirm)
+```
+
+All accept `-p/--stardag-profile` and `-e/--stardag-env` to target a
+non-active profile / environment.
+
 ### Modal Integration
 
 ```bash
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,12 +11,13 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 - **CLI: `stardag concurrency-limits` command group for managing named
   concurrency limits.** Wraps the registry's concurrency-limit endpoints for
   the active profile / environment (override with `-p/--stardag-profile` and
-  `-e/--stardag-env`): `list` (with optional `--holders` counts), `set <key>
-<max_concurrent>` (upsert), `delete <key>` (`--yes` to skip confirmation),
-  `holders <key>` (RUNNING slot holders, oldest first), and `evict <key>
-<task_id>` to free leaked slots. Replaces the need for an ad-hoc script to
-  `PUT /api/v1/concurrency-limits/{key}`. Backed by new `APIRegistry`
-  `concurrency_limit_{list,set,delete,holders,evict}` methods.
+  `-e/--stardag-env`). Subcommands: `list` (with optional `--holders` counts);
+  `set` to upsert a limit (`stardag concurrency-limits set <key> <max_concurrent>`);
+  `delete <key>` (`--yes` to skip confirmation); `holders <key>` (RUNNING slot
+  holders, oldest first); and `evict` to free leaked slots
+  (`stardag concurrency-limits evict <key> <task_id>`). Replaces the need for an
+  ad-hoc script to `PUT /api/v1/concurrency-limits/{key}`. Backed by new
+  `APIRegistry` `concurrency_limit_{list,set,delete,holders,evict}` methods.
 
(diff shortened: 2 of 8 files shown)


andhus:
Reviewed. Solid, convention-consistent CLI addition — profile save/restore matches the `modal.py` precedent exactly, all five client methods reuse `_request` (rate-limit retry + typed error mapping), `registry.close()` in `finally` everywhere, confirm-declined is tested for **both** delete and evict with `assert_not_called()`, server-side 403s surface cleanly, and docs are complete per repo policy (CLI reference, CHANGELOG, skill bundle). Nothing blocks merge; three small robustness fixes worth taking:

**minor — user-supplied `key` (and `task_id`) are f-string-interpolated into the URL path unencoded.** A key containing `/`, `?`, `#`, `%` or space breaks routing or mangles the query string — and the server accepts such keys when created through the UI (which percent-encodes them), so a UI-created key like `a/b` is unmanageable and undeletable from the CLI (`set`/`delete` 404). `urllib.parse.quote(key, safe="")` on the path segment, mirroring the UI client. This is the one finding with a data-model implication.

**minor — transport errors dump raw tracebacks.** Only `StardagError` is caught; `httpx.ConnectError`/`ReadTimeout` (registry unreachable, VPN down, wrong URL) propagate as a multi-screen traceback instead of `Error: …` + exit 1. Catch `httpx.HTTPError` alongside, or wrap transport errors in `_request`.

**minor — `-e/--stardag-env` is silently ignored under API-key auth.** The resolved `environment_id` is only transmitted for JWT auth, and API keys are server-scoped to their own environment — `list -e staging` with an API-key profile silently returns the key's environment instead. Warn or error when `-e` is passed with API-key auth.

**nits:** the evict confirm *prompt* says only "This records TASK_FAILED for it." — the "only evict holders whose process you know is dead" oversubscription warning made it into `--help` but not the interactive prompt, which is where the decision happens; old-server 404s surface cleanly but a "server may not support concurrency limits (requires a newer stardag-api)" hint would save a debugging session; `limits_set` uses direct `result["key"]` indexing where the rest of the file uses defensive `.get(...)`.

**test gaps:** special-character key (would have caught the encoding finding); transport-error handling; `-e`-under-API-key behavior.

andhus:
Addressed the review feedback in 4a73bfc:

**1. URL-encode `key` / `task_id` path segments** (human reviewer + Copilot)
Each embedded path segment in the `APIRegistry.concurrency_limit_{set,delete,holders,evict}` methods is now percent-encoded with `urllib.parse.quote(seg, safe="")`. A key like `a/b` (or one containing other reserved characters) stays a single path segment instead of splitting the route / hitting the wrong endpoint. Added `TestConcurrencyLimitPathEncoding` in `test_api_registry.py` asserting the encoded wire path (via `url.raw_path`, since `url.path` decodes for display) for all four methods, including a key with `/`, space, and `#`, plus the `task_id` segment in `evict`.

**2. Accept a raw environment UUID for `-e/--stardag-env`** (Copilot)
`_resolve_registry` now detects when the value is already a UUID (`_looks_like_uuid`) and uses it directly, skipping the slug->id resolution that requires registry/workspace context — so passing a raw environment UUID works without a fully-configured profile, matching the documented "slug or ID" behavior. Slug resolution is unchanged for non-UUID values (and still errors clearly when context is missing). Added tests in `TestResolveEnv` covering: UUID used directly, UUID working without registry/workspace context, slug resolved via lookup, and slug still erroring without context.

**3. CHANGELOG inline-code formatting** (Copilot)
Rewrote the concurrency-limits entry so every backticked command example sits on one line (e.g. `` `stardag concurrency-limits set <key> <max_concurrent>` ``); no inline-code span is broken across a newline anymore. The other flagged class of issue elsewhere in the file lives only in an already-released section, so it's left untouched to avoid unrelated churn.

Tests: full non-`modal_live` SDK suite green (821 passed, 6 skipped); pre-commit (ruff, ruff-format, pyright, prettier) clean on the changed files.

PR #178, opened 2026-07-15
UI: hierarchical breadcrumb for the task Execution section

andhus:
## What

Redesigns the **Execution** section of the task detail panel (`app/stardag-ui`). The previous flat label/value rows are replaced with a hierarchical breadcrumb, and the collapsible "More details" block becomes a grouped 2-column table.

### Before / after (Execution section)

**Before** — a stack of label/value rows:

```
⚡ Modal
App:              my-app
Function:         worker_default
Call ref:         fc-…            [copy]
Workspace / env:  my-workspace / staging
▸ More details
```

**After** — one hierarchical breadcrumb:

```
⚡ Modal
my-workspace / staging / my-app / worker_default / fc-…   [copy]
▸ More details
```

## Breadcrumb behavior

- Four top-level segments: `[workspace / environment] / app_name / function_name / <call-ref>`.
  - Workspace and environment **collapse into one linked segment** — a bare `modal.com/apps/{workspace}` URL is misleading (it resolves to whichever environment was "latest active"), so the coarsest addressable level is the workspace+environment pair.
- Each segment is a **best-effort Modal dashboard deep link** (`target="_blank" rel="noopener noreferrer"`); when the ids needed to build that level's URL are missing, the segment renders as **plain text** but still shows the label.
- **Absent segments are omitted** — the trail starts at the first level actually recorded. Renders nothing for an explicitly non-modal `kind`, or when there's nothing to show.
- The call-ref segment keeps its **click-to-copy** button.

### Wrapping / hanging indent

Each top-level segment plus its trailing ` /` is one `whitespace-nowrap` unit; the only breakable whitespace sits *between* units. So a line can wrap **only right after a slash** — never mid-segment, never before the slash (the workspace/environment pair stays glued). The block uses `pl-4 -indent-4` for a hanging indent: the first line is flush, wrapped continuation lines are indented ~1rem. Separators are muted.

## URL builders (`utils/modalLinks.ts`)

New, matching the existing style (all `encodeURIComponent`-safe, `string | null`, graceful fallback):

- `modalEnvironmentUrl(m)` → `…/apps/{workspace}/{environment}` (null if either missing; no "main" fallback here).
- `modalFunctionUrl(m)` → the app URL + `?activeTab=functions&functionId={function_id}` (null without `function_id` or a resolvable app URL).

Reused alongside the existing `modalAppUrl` / `modalFunctionCallUrl`.

## "More details" table

Grouped top-down: human-readable names (Workspace, Environment, App, Function), a hairline divider, then raw ids (App ID, Function ID, Call ref). The redundant `kind` row is dropped; the divider shows only when both groups have fields; only present fields render; each value keeps click-to-copy.

## Tests

Full UI suite green: **84 tests** across 8 files. New/updated coverage for the URL builders (happy path, null-on-missing-field, reserved-char encoding) and the breadcrumb (order, omitted/plain-text segments, combined ws/env edge cases, call-ref copy, non-modal gating) plus the grouped table.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,23 @@ All notable changes to the Stardag project (SDK, Registry API, and UI).
 
 For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
+## [Unreleased]
+
+### UI
+
+- **Task detail "Execution" section is now a hierarchical breadcrumb.** The
+  previous flat label/value rows (App, Function, Call ref, Workspace / env)
+  are replaced by a single `workspace / environment / app_name /
+function_name / <call-ref>` trail (workspace and environment collapse into
+  one linked segment). Each present level is a best-effort deep link into the
+  corresponding Modal dashboard page (opening in a new tab), falling back to
+  plain text when the ids needed to build that level's URL are absent; absent
+  levels are omitted so the trail starts at the first level actually
+  recorded. The call-ref segment keeps its click-to-copy button. The
+  collapsible "More details" block now renders the captured identifiers as a
+  2-column table (human-readable names, then a divider, then the raw ids),
+  each value click-to-copy.
+
 ## [0.11.0] — 2026-07-15
 
 ### SDK
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,16 +10,15 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
 - **Task detail "Execution" section is now a hierarchical breadcrumb.** The
   previous flat label/value rows (App, Function, Call ref, Workspace / env)
-  are replaced by a single `workspace / environment / app_name /
-function_name / <call-ref>` trail (workspace and environment collapse into
-  one linked segment). Each present level is a best-effort deep link into the
-  corresponding Modal dashboard page (opening in a new tab), falling back to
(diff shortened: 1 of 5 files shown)


andhus:
Addressed both Copilot findings in 39df433:

**1. Call-ref linked to the wrong level when `function_id` is missing (bug).**
`modalFunctionCallUrl` intentionally falls back to the app page when `function_id` is absent (other callers, e.g. the "View on Modal" links, rely on that shared fallback), so a clickable call ref could navigate to the coarser app level. Fixed in the breadcrumb only: the call-ref link is now gated on `modalFunctionUrl(metadata)` being non-null (genuine `function_id` + resolvable app URL) — `const callUrl = modalFunctionUrl(metadata) ? modalFunctionCallUrl(metadata, executorRef) : null;`. When the gate is false the ref renders as plain text and keeps its copy button. `modalFunctionCallUrl` is unchanged. Added a test: `app_name`/`app_id` + `executorRef` but no `function_id` → call ref is plain text (not an anchor) with the copy button still working.

**2. CHANGELOG Unreleased bullet hard-wrapped an inline-code path to column 0.**
Reworded to use a short `workspace / environment` code span plus separate `app_name` / `function_name` spans instead of one long path, so no inline-code span is long enough to wrap. `prettier` now leaves every continuation line properly indented under the bullet (confirmed clean, no reformat).

UI suite: 86 passed (8 files). `uvx pre-commit run` (prettier / eslint / tsc): all pass.

Issue #180, opened 2026-07-15
Unit tests are affected by local active stardag config (target-root config leaks into `.target()` tests)


andhus:
## Summary

Some unit tests read the developer's **ambient/active stardag configuration**
instead of a hermetic test setup. A developer whose active config sets a
**non-local default target root** (any non-filesystem scheme — e.g.
`s3://my-bucket/roots/default`) sees these tests fail, even though the tests
have nothing to do with that config.

Tests should never depend on local active config. All tests that touch targets
should go through the "targets test harness"-style setup (in-memory / tmp-path
target factory) so the result is independent of the machine's config.

## Observed failures

With a non-local default target root configured locally, these fail:

- `tests/test__core/test_task.py::TestTarget::test_target_returns_serializable`
- `tests/test__core/test_task.py::TestTarget::test_target_has_correct_serializer`
- `tests/test__core/test_task.py::TestTarget::test_target_uri_matches_relpath`
- `tests/test__core/test_task.py::TestDirectoryTarget::test_target_returns_directory_serializable`

with, e.g.:

```
ValueError: URI s3://my-bucket/roots/default/... does not match any of the
configured prefixes: ['/'].
```

(There may be other target-touching tests with the same latent dependency — an
audit is warranted, not just these four.)

## Reproduction

1. Configure any non-local default target root in your **user-level config
   file / active profile** (not via env var), e.g. a `default` root of
   `s3://my-bucket/...`.
2. Run the unit test suite.
3. The tests above fail while resolving `task.target()`.

## Root cause

Two independent gaps combine:

1. **The affected tests don't isolate the target factory.** They call
   `task.target()` without requesting a target-isolation fixture, so
   `target()` falls through to `target_factory_provider.get()`, which builds a
   factory from the **ambient config**. Compare `TestRunAndSave`, which takes
   the `default_in_memory_fs_target` fixture and is therefore hermetic.

2. **The isolation that does exist only covers env vars, not config files.**
   The autouse `cleared_stardag_env_vars` fixture (in
   `lib/stardag/tests/conftest.py`) clears `STARDAG_*` environment variables,
   but config resolved from a **user-level config file / active profile**
   survives it. So even with all `STARDAG_*` vars cleared, the active profile's
   target roots still leak into tests.

## Suggested fix

- **Harness-level (preferred):** make config hermetic for the whole unit suite,
  not just env vars. An autouse fixture should reset the config to a known,
  in-repo default independent of user config files, active profile, `cwd`
[...]

Issue #153, opened 2026-06-25
Naked `: BaseTask` field annotation silently drops subclass params and crashes on load — validate at class-construction time?


andhus:
## Summary

Annotating a task field with the **abstract** task base directly — a "naked"
`child: BaseTask` (or `child: Task`) — rather than wrapping it in
`SubClass[...]` / `Polymorphic()` appears to never produce desired behaviour,
yet it is silently accepted at class-definition time. It then fails in two
ways that are easy to miss until much later:

1. **Lossy serialization.** `model_dump` emits only the base-class fields and
   the polymorphic discriminator, **silently dropping every subclass-specific
   parameter**. The dropped data is unrecoverable, and because the task id is
   derived from this dump, the id itself bakes in the lossy form.
2. **Crash on load.** Round-tripping the dump back through `model_validate`
   skips polymorphic dispatch and tries to instantiate the abstract base
   directly: `TypeError: Can't instantiate abstract class BaseTask without an
   implementation for abstract method 'complete'`.

The dangerous part is that (1) is completely silent. A field declared this way
works fine for as long as nobody assigns a value to it; the first task that
actually populates the field gets serialized lossily into the registry, and
the corruption only surfaces later when something tries to read it back.

Tested on `stardag==0.8.1`.

## Minimal reproduction

```python
import stardag as sdag


class Child(sdag.Task[str]):
    payload: str

    def run(self) -> None:
        self.get_target().save(self.payload)


class ParentBare(sdag.Task[str]):
    # The problem: annotated with the abstract task base directly,
    # not wrapped in sdag.SubClass[...] / Polymorphic().
    child: sdag.BaseTask

    def run(self) -> None:
        self.get_target().save("ok")


parent = ParentBare(child=Child(payload="important-config"))

# 1) Lossy serialization — `payload` is silently dropped:
dumped = parent.model_dump(mode="json")
print(dumped["child"])
# -> {'__namespace': '', '__name': 'Child', 'version': ''}

# 2) Crash on load:
ParentBare.model_validate(dumped)
# TypeError: Can't instantiate abstract class BaseTask without an
#            implementation for abstract method 'complete'
```

The working form is to wrap the base in `SubClass`:

```python
class ParentSub(sdag.Task[str]):
    child: sdag.SubClass[sdag.BaseTask]  # == Annotated[BaseTask, Polymorphic()]
    ...
```

With this, `model_dump` preserves the full child body
[...]

PR #184, opened 2026-07-15
Enforce strict polymorphic exact-type on the deserialize path too

### LINKED ISSUE #181: Suggestion: also enforce exact-type for strict concrete polymorphic fields on the dict/deserialize path
## Context

PR #179 (closing #153) makes a bare **concrete** polymorphic field
(e.g. `child: MyTask`, without `SubClass[...]`) a *strict* field: it means
exactly `MyTask`. Passing a **subclass instance** now raises
`StrictPolymorphicTypeError` at construction, because otherwise the subclass's
extra parameters are silently dropped on serialization and task identities
collide.

That enforcement runs in a `model_validator(mode="after")`, so it covers the
**instance path** (`Parent(child=ChildOfMyTask(...))`) — the natural way this
mistake is made.

## Suggestion (optional tightening)

The **dict / deserialize path** is not covered:

```python
Parent.model_validate({"child": {"__name": "ChildOfMyTask", ...}})
```

Here pydantic coerces the child-dict straight into the base `MyTask` (ignoring
the discriminator, since a strict field isn't polymorphic), producing a
truncated `MyTask` instance. By the time the after-validator runs, the value is
already exactly `MyTask`, so nothing is flagged.

This is currently a **non-issue in practice**: with the PR merged, stardag's own
serialization can no longer *produce* such a payload (you can't construct the
offending instance to dump it). It can only arise from:

- hand-crafted / externally-supplied input dicts, or
- data serialized by a **pre-fix** library version.

We could optionally close it with a `mode="before"` validator on strict fields
that inspects the raw value: if it's a dict whose `__namespace`/`__name`
discriminator resolves to a class other than the declared strict type, raise
(mirroring the instance-path error).

## Trade-off to decide

Doing this would make **loading pre-fix data that was already lossily truncated
into a strict field fail loudly** instead of loading the (already-degraded)
base type as it does today. That is arguably the correct "fail loud on corrupt
data" behavior, but it is a behavior change for existing stored payloads, so it
warrants a deliberate decision rather than shipping silently.

Filing as a reference/option, not a bug — the instance-path enforcement already
covers the realistic case.

andhus:
Closes #181.

## Context

v0.12.0 (#179) made a bare **concrete** polymorphic field (`child: MyTask`,
without `SubClass[...]`) *strict* — it means exactly `MyTask`. Assigning a
subclass **instance** raises `StrictPolymorphicTypeError`. But that enforcement
is an after-validator, so it only covers the **instance path**.

## Problem

A strict field is non-polymorphic, so the **deserialize path** was still a silent
trap:

```python
Parent.model_validate({"child": {"__name": "ChildOfMyTask", ...}})
```

pydantic coerced the child-dict straight into the base `MyTask` (ignoring the
discriminator), silently dropping the subclass's parameters — the very
truncation #153 is about.

## Fix

Add a `mode="before"` validator that inspects the **raw** input: a dict whose
`__namespace`/`__name` discriminator resolves to a class other than the declared
strict type is rejected with the same `StrictPolymorphicTypeError`. Plain dicts
without a discriminator are untouched (validated as the exact strict type — no
false positives).

The annotation/value walk is refactored to take a leaf-check callback, so the
instance path (after-validator) and the discriminator-dict path (before-validator)
share one traversal — no duplicated container/union/`Annotated` recursion.

```python
class Parent(sd.Task[int]):
    child: MyTask                                   # strict

Parent.model_validate({"child": exact_MyTask_dump})        # OK
Parent.model_validate({"child": {"a": 1}})                 # OK (plain dict → exact type)
Parent.model_validate({"child": subclass_discriminator})   # StrictPolymorphicTypeError
```

## Behavior-change note

Loading data that a **pre-0.12.0** version already truncated into a strict field
now raises instead of loading the degraded base type — the correct "fail loud on
corrupt data" behavior (the trade-off called out in #181). Switch the field to
`SubClass[...]` if it should accept subclasses.

## Verification

- `ruff` / `ruff-format` / `pyright` clean on changed files.
- Full suite green except the 4 pre-existing environment-local target-root
  failures (unrelated; fixed separately in the hermetic-tests PR / #180).
- Adds `TestStrictConcretePolymorphicDeserialize` (exact-type dict accepted,
  plain dict accepted, subclass-discriminator dict / in-list / in-Optional
  rejected). CHANGELOG `[Unreleased]` entry added.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ All notable changes to the Stardag project (SDK, Registry API, and UI).
 
 For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
-## [Unreleased]
+## [0.13.0] — 2026-07-16
 
 ### SDK
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,22 @@ All notable changes to the Stardag project (SDK, Registry API, and UI).
 
 For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
+## [Unreleased]
+
+### SDK
+
+- **Strict (bare concrete) polymorphic fields now also reject subclass data on
+  the deserialize path.** v0.12.0 rejected a subclass _instance_ assigned to a
+  strict field; it now also rejects _serialized data_ — an input dict whose
+  `__namespace`/`__name` discriminator resolves to a subclass of the declared
+  strict type — instead of silently coercing it into the base type (dropping the
+  subclass's parameters). Plain dicts without a discriminator are unaffected
+  (validated as the exact strict type). **Note:** loading data that was already
+  lossily truncated into a strict field by a pre-0.12.0 version now raises
+  `StrictPolymorphicTypeError` rather than loading the degraded base type — the
+  correct "fail loud on corrupt data" behavior; switch the field to
+  `SubClass[...]` if it should accept subclasses.
+
 ## UI Only — 2026-07-15
 
 ### UI
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,6 +6,41 @@ For changes to the Registry API, UI, and other components, see [CHANGELOG.md](CH
 
 ---
 
+## v0.13.0 — Strict polymorphic fields: also enforced on load
+
+Follow-up to v0.12.0. A _strict_ polymorphic field — a bare concrete
+annotation like `dep: MyTask` (no `SubClass[...]`) — means exactly `MyTask`.
+v0.12.0 rejected a subclass **instance** assigned to such a field; v0.13.0
+closes the remaining gap on the **deserialize path**. Run `pip install -U stardag`.
+
+Previously, validating serialized data (e.g. via `model_validate` /
+registry rehydration) with a subclass payload at a strict field silently
+coerced it into the base type, dropping the subclass's parameters. Now an
+input dict whose `__namespace`/`__name` discriminator names anything other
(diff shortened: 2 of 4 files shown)


andhus:
Addressed the review (both comments — the family-root gap):

- **`_strict_discriminator_leaf`**: now requires **both** `__namespace` and `__name` to treat a payload as a discriminator (matching `Polymorphic.dispatch`). Crucially, when the declared strict type is a **family root** (a direct `PolymorphicRoot` child — unregistered, no `__type_id__`, and never serializable with a discriminator itself), *any* discriminator payload is now rejected, since one necessarily names a subclass. Previously such fields skipped the check entirely and silently truncated.
- **Test**: added `test_family_root_strict_type_rejects_subclass_discriminator` covering `field: FamilyRoot` — a plain dict coerces to the exact root, a subclass discriminator payload raises `StrictPolymorphicTypeError`.

Also rebased onto latest `main` (picks up #183, the hermetic-test fix). Full suite green (673 passed); ruff/ruff-format/pyright clean.

andhus:
Prepared this for release as **v0.13.0** (minor, not a patch): it changes behavior for previously-non-erroring loads — data a pre-0.12.0 version already truncated into a strict field now fails loudly instead of silently loading the degraded base type. Per the 0.x convention used for 0.12.0, a behavior change goes in the minor slot.

- `CHANGELOG.md`: promoted `[Unreleased]` → `[0.13.0] — 2026-07-16`.
- `RELEASE_NOTES.md`: added the v0.13.0 SDK entry with a ⚠️ migration guide.

On merge, the release follows the documented procedure (tag `v0.13.0` → CI publishes to PyPI → refine the GitHub Release).

PR #193, opened 2026-07-17
self-host: complete setup — dedicated Modal env, primary workspace mapping, one-command DAG-execution hookup

andhus:
> **Stacked PR**: based on `server-release-pipeline` (#191). Only the last three commits are new here.

## Motivation

`stardag self-host up` previously left you with a bare server: you still had to add a registry entry, log in, create a workspace/environment, mint an API key, push it to Modal, and configure a target root before the first DAG could run. This PR makes `up` finish the job, mirroring Modal's own workspace/environment model:

- **Shared Modal workspace** (team account): you get a shared Stardag workspace with the same name — teammates who sign in land in the same place.
- **Personal Modal workspace**: your personal Stardag workspace is used — no artificial shared workspace.
- Either way, a **`main` Stardag environment** mirrors Modal's default environment, and DAG execution on Modal is wired up with one API key.

## What the defaults create

| What | Default | Purpose |
| --- | --- | --- |
| Modal environment `stardag-server` | server app + `-config`/`-jwt` secrets + `-meta` Dict | Isolates the server from the Modal environments where DAG apps run |
| Stardag workspace | named after the shared Modal workspace, or the personal workspace | Mirrors the Modal account structure |
| Stardag environment | `main` | Mirrors Modal's default environment |
| API key → Modal secret `stardag-api-key` | pushed into the DAG-execution Modal environment (account default) | Modal-executed DAGs authenticate against the registry |
| Target root `default` | `modalvol://stardag/<workspace-slug>` | Task outputs on a Modal volume |
| Local registry + profile `selfhosted` | `~/.stardag/config.toml` | SDK/CLI ready immediately |

The connect phase is also a standalone, idempotent command: `stardag self-host connect` (re-runs, `--skip-connect` recovery, OIDC mode).

## Server-side bootstrap (local auth mode)

New `AUTH_PRIMARY_WORKSPACE_NAME` / `AUTH_PRIMARY_WORKSPACE_ENVIRONMENT` (default `main`, empty disables) settings: startup idempotently ensures the named shared workspace (bootstrap admin as owner) + environment — or the environment in the admin's personal workspace when no name is set. The CLI bakes these into the config secret so the server provisions them on first boot. No-op in OIDC mode.

## Flags

| Flag | Commands | Default |
| --- | --- | --- |
| `--server-modal-env` | up/connect/upgrade/status/destroy | `stardag-server` (created on demand; `''` = Modal default env, for pre-existing deployments) |
| `--primary-workspace` / `--no-primary-workspace` | up/connect | shared Modal workspace name / skipped for personal accounts |
| `--execution-modal-env` | up/connect | account's default Modal environment |
| `--target-root name=uri` / `--no-target-root` | up/connect | `default=modalvol://stardag/<workspace-slug>` |
| `--registry-name` / `--profile-name` | up/connect | `selfhosted` |
| `--skip-connect` | up | off |

Every prompt has a flag equivalent; `--yes` takes all defaults.

## OIDC mode

Server-side bootstrap is local-mode-only, so `up --auth-mode oidc` prints a pointer to `stardag self-host connect`, which runs the same browser PKCE login as `stardag auth login` and then provisions workspace/environment/API key/target root/profile via the API (the workspace-create and `/auth/exchange` endpoints accept the login token directly).

## Tests

- `app/stardag-api`: 358 passed, 18 skipped — includes new `test_primary_workspace.py` (shared-name workspace + owner + `main` env, personal-workspace env, idempotency on restart, no demotion of adjusted memberships, OIDC/unset no-ops, empty-env disable).
- `lib/stardag`: 751 passed (full unit suite) — new `test_selfhost/test_complete_setup.py` covers primary-workspace derivation (shared/personal/explicit/opt-out/interactive), `environment_name` plumbing for secrets/Dict/`build_server_app` (mocked Modal), and the connect flow end-to-end against a mocked registry API (`httpx.MockTransport`, isolated home dir): personal + created + existing-idempotent workspace paths, missing-env creation, target-root override/skip/keep-existing, non-fatal Modal-secret push failure.
- `docs`: `mkdocs build` clean.

## Notes

- `--execution-modal-env` defaults to *Modal's default environment* (push with `environment_name=None`) rather than hard-coding `main` — same result for standard accounts, but robust for workspaces whose default env is named differently.
- `upgrade`/`status`/`destroy` against deployments created by older SDK versions (whose objects live in the default Modal env) work via `--server-modal-env ''`; the `upgrade` missing-secret error says so.

https://claude.ai/code/session_01ARrhAGii9mMcpyVNTbSMEf

### DIFF
--- a/.claude/skills/stardag/registry-and-platform.md
+++ b/.claude/skills/stardag/registry-and-platform.md
@@ -175,15 +175,31 @@ stardag modal stardag-api-key create        # Create API key for Modal
 
 ### Self-Hosting (API + UI on Modal)
 
-Requires the `selfhost` extra (`pip install "stardag[selfhost]"`) and a
-stardag repo checkout. See docs/how-to/self-host-modal.md.
+Requires the `selfhost` extra (`pip install "stardag[selfhost]"`). Deploys
+the prebuilt server image by default (`--from-source` builds from a repo
+checkout). See docs/how-to/self-host-modal.md.
 
 ```bash
-stardag self-host up        # Provision Neon DB, migrate, deploy API+UI, print URL
-stardag self-host upgrade   # Apply migrations + redeploy from current source
+stardag self-host up        # Provision Neon DB, migrate, deploy API+UI, complete setup
+stardag self-host connect   # (Re)run the post-deploy setup only (idempotent)
+stardag self-host upgrade   # Apply migrations + redeploy
 stardag self-host status    # Deployment status + URL
 stardag self-host destroy   # Stop the Modal app (DB untouched)
 ```
+
+The server app + its secrets live in a dedicated Modal environment
+(default `stardag-server`, flag `--server-modal-env`), isolated from the
+environments where user DAG apps run. `up`/`connect` also complete the
+setup: a primary Stardag workspace (named after the shared Modal workspace,
+or the personal workspace; `--primary-workspace`/`--no-primary-workspace`),
+a `main` environment, an API key pushed as Modal secret `stardag-api-key`
+into the DAG-execution Modal environment (`--execution-modal-env`, default:
+the account's default env), a default target root
+`modalvol://stardag/<workspace-slug>` (`--target-root`/`--no-target-root`),
+and a local SDK registry + profile named `selfhosted`
+(`--registry-name`/`--profile-name`). In OIDC auth mode run
+`stardag self-host connect` after `up` (browser login, then the same setup
+via the API).
 
 ## Configuration System
 
--- a/.claude/skills/stardag/registry-and-platform.md
+++ b/.claude/skills/stardag/registry-and-platform.md
@@ -187,19 +187,20 @@ stardag self-host status    # Deployment status + URL
 stardag self-host destroy   # Stop the Modal app (DB untouched)
 ```
 
-The server app + its secrets live in a dedicated Modal environment
-(default `stardag-server`, flag `--server-modal-env`), isolated from the
-environments where user DAG apps run. `up`/`connect` also complete the
-setup: a primary Stardag workspace (named after the shared Modal workspace,
-or the personal workspace; `--primary-workspace`/`--no-primary-workspace`),
-a `main` environment, an API key pushed as Modal secret `stardag-api-key`
-into the DAG-execution Modal environment (`--execution-modal-env`, default:
-the account's default env), a default target root
-`modalvol://stardag/<workspace-slug>` (`--target-root`/`--no-target-root`),
-and a local SDK registry + profile named `selfhosted`
-(`--registry-name`/`--profile-name`). In OIDC auth mode run
-`stardag self-host connect` after `up` (browser login, then the same setup
-via the API).
+The server app (default name `server`) + its secrets live in a dedicated
+Modal environment (default `stardag-host`, flag `--server-modal-env`),
+isolated from the environments where user DAG apps run; the default URL is
+`https://<workspace>-stardag-host--server.modal.run`. `up`/`connect` also
+complete the setup: a primary Stardag workspace (named after the shared
+Modal workspace, or the personal workspace;
+`--primary-workspace`/`--no-primary-workspace`), a `main` environment, an
+API key pushed as Modal secret `stardag-api-key` into the DAG-execution
+Modal environment (`--execution-modal-env`, default: the account's default
+env), a default target root `modalvol://stardag-targets/<workspace-slug>`
+(`--target-root`/`--no-target-root`), and a local SDK registry + profile
+named `selfhosted` (`--registry-name`/`--profile-name`). In OIDC auth mode
+run `stardag self-host connect` after `up` (browser login, then the same
+setup via the API).
 
 ## Configuration System
 
--- a/.claude/skills/stardag/registry-and-platform.md
+++ b/.claude/skills/stardag/registry-and-platform.md
@@ -196,7 +196,8 @@ Modal workspace, or the personal workspace;
 `--primary-workspace`/`--no-primary-workspace`), a `main` environment, an
 API key pushed as Modal secret `stardag-api-key` into the DAG-execution
 Modal environment (`--execution-modal-env`, default: the account's default
-env), a default target root `modalvol://stardag-targets/<workspace-slug>`
+env), a default target root
+`modalvol://stardag-targets-<workspace-slug>-<environment-slug>/default`
 (`--target-root`/`--no-target-root`), and a local SDK registry + profile
 named `selfhosted` (`--registry-name`/`--profile-name`). In OIDC auth mode
 run `stardag self-host connect` after `up` (browser login, then the same
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,8 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
   workspace) with a `main` environment; an API key is minted and pushed
   as the Modal secret `stardag-api-key` into the DAG-execution Modal
   environment (`--execution-modal-env`); a default target root
-  `modalvol://stardag-targets/<workspace-slug>` is registered
+  `modalvol://stardag-targets-<workspace-slug>-<environment-slug>/default`
+  (a dedicated Modal volume per workspace + environment) is registered
   (`--target-root`/`--no-target-root`); and a local SDK registry +
   profile (`selfhosted`) are written. In OIDC auth mode, `connect` runs
   the browser login first and provisions via the API.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,19 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
   (`--server-version`); `--from-source` builds from a repo checkout
   (the UI compiles inside the Modal image build — no local Node/Docker
   required). See the "Self-host on Modal" guide.
+- **`stardag self-host up` completes the whole setup** (new `connect`
+  subcommand re-runs it idempotently): the server app + its secrets are
+  isolated in a dedicated Modal environment (`stardag-server`, flag
+  `--server-modal-env`); a primary Stardag workspace is created mirroring
+  a shared Modal workspace's name (`--primary-workspace` /
+  `--no-primary-workspace`; personal Modal accounts use the personal
+  workspace) with a `main` environment; an API key is minted and pushed
+  as the Modal secret `stardag-api-key` into the DAG-execution Modal
+  environment (`--execution-modal-env`); a default target root
+  `modalvol://stardag/<workspace-slug>` is registered
+  (`--target-root`/`--no-target-root`); and a local SDK registry +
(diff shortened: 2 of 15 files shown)


andhus (commented):
Reviewed the branch HEAD against its base (`server-release-pipeline`), i.e. only this PR's three commits. This is a well-executed completion of the self-host story, and it also lands clean resolutions to several items from the earlier review rounds on the stack: keep-warm is now persisted in the app's meta Dict, `upgrade` defaults to the recorded deployed version with an explicit-downgrade warning, the CLI now mirrors the server's full password policy (8-char min *and* 72-byte max), and the config-flag fail-fast guard is present in the rewritten `up()`. The Modal `environment_name` plumbing is consistent end to end (verified: config/JWT secret push + existence checks, `Secret.from_name` inside `build_server_app`, `app.run()` for migrate, `deploy()`, web-URL lookups, meta Dict, `status`, `connect`, and `destroy` incl. `--env` on app stop and Dict/secret deletion — no object I could find still lands in the default env unless `--server-modal-env ''` is chosen, and the execution-env secret intentionally goes elsewhere). The server-side bootstrap is restart-safe in the sequential sense (match-before-create everywhere, never renames/demotes/deletes), the connect flow is genuinely re-runnable, and test coverage of both sides is thorough. Findings:

**Finding 1 (severity: medium)** — `lib/stardag/src/stardag/_cli/_selfhost_connect.py:382` — the connect flow is idempotent for workspace/environment/target-root, but the API key is minted unconditionally on every run: `_create_stardag_api_key` always POSTs a new key, the server accepts duplicate names, and previous keys are never revoked. Every `connect` re-run (the advertised recovery path for `--skip-connect`, push failures, etc.) therefore leaves an additional *live, untracked* credential in the registry while only the newest lands in the Modal secret — contradicting the module docstring's "Everything is idempotent" and steadily worsening credential hygiene. Suggested fix: list the environment's keys first and revoke the same-named key before creating (delete+create, mirroring `_push_modal_secret`'s semantics), or at minimum print a warning that a previously created key remains active and should be revoked in the UI.

**Finding 2 (severity: low)** — `app/stardag-api/src/stardag_api/services/local_auth.py:300-317` — `_ensure_shared_workspace` adopts *any* non-personal workspace whose name matches `AUTH_PRIMARY_WORKSPACE_NAME` exactly. In a deployment with `--enable-registration`, a regular user can pre-create a workspace with the (predictable — it's the Modal workspace name) primary name; on the next restart the bootstrap admin is silently added as OWNER of that user-controlled workspace, and it becomes the deployment's primary — the connect flow will then create API keys and target roots inside a workspace another user co-owns and can inspect/manage. Suggested fix: scope the match to workspaces created by the bootstrap admin (`created_by_id == admin.id`), or persist the bootstrapped workspace id (e.g. alongside the meta Dict) instead of re-matching by name.

**Finding 3 (severity: low)** — `app/stardag-api/src/stardag_api/services/local_auth.py` (`ensure_bootstrap_admin` / `ensure_primary_workspace`) — the startup bootstrap is idempotent across sequential restarts but not concurrency-safe: the combined server image's default `CMD` runs `GUNICORN_WORKERS=2`, and each worker executes the lifespan. On a first cold start two workers can pass the "not found" check simultaneously; the loser hits the unique constraint (user `external_id`/email, workspace slug) and the resulting `ValueError`/`IntegrityError` propagates out of the lifespan and kills that worker's boot (self-heals on respawn, but presents as a startup crash on exactly the first launch — the worst moment for a self-hoster). Modal's single-process ASGI runtime dodges this, but ECS/docker deployments of the same image with local auth mode won't. Suggested fix: catch `IntegrityError` (and the wrapped `ValueError` from `create_local_user`) in the ensure-functions and re-fetch — "another worker won" is success.

**Finding 4 (severity: low)** — `lib/stardag/src/stardag/_cli/_selfhost_connect.py:382` + `lib/stardag/src/stardag/_cli/modal.py:201-236` — `_create_stardag_api_key` was written for the `stardag modal stardag-api-key create` command: on 401 it prints "Try running: stardag auth refresh" and raises `typer.Exit`. Inside `run_connect` that advice is wrong (the connect flow just minted the token itself), and the `Exit` aborts after workspace/environment/target-root are ensured but before the local SDK registry/profile are written — recoverable by re-running, but the user isn't told so. Suggested fix: wrap the call in connect-specific error handling (like the Modal-secret push already has) with a "re-run `stardag self-host connect`" pointer.

https://claude.ai/code/session_01ARrhAGii9mMcpyVNTbSMEf

andhus:
## Review findings addressed

| Finding | Resolution | Commit |
| --- | --- | --- |
| **1 (medium)** — connect mints a new API key on every run | Key handling is now delete+create like the Modal-secret push: `_rotate_api_key` lists the environment's keys, revokes any *live* key with the same name (printed: `Revoked previous API key "<name>" (prefix: ...)`), then mints a fresh one. Already-revoked and differently-named keys are untouched. Module docstring updated to state the rotate semantics. Tests: `test_connect_rotates_existing_api_key` (pre-existing live key revoked; two consecutive runs leave exactly one live key). | 497a7fc |
| **2 (low)** — primary-workspace adoption of a user-controlled workspace | `_ensure_shared_workspace` now only matches workspaces with `created_by_id == admin.id`. A same-named workspace the admin didn't create is logged (`not adopting it - creating a separate primary workspace instead`) and a fresh admin-owned workspace is created, its slug getting the existing random-suffix collision handling. Test: `test_primary_workspace_not_adopted_from_other_user` (attacker-pre-created workspace gets no admin membership; the new admin-owned workspace is used and re-runs adopt it idempotently). | 220501c |
| **3 (low)** — multi-worker bootstrap race crashes a worker's first boot | `ensure_bootstrap_admin` catches the duplicate-email `ValueError` from `create_local_user` (which already rolled back) and logs "created concurrently by another worker". `ensure_primary_workspace` wraps its provisioning pass in an `IntegrityError` handler: rollback, log, and re-run the match-before-create pass, which then finds the rows the other worker created (mirrors the `get_or_create_user_from_oidc_claims` race pattern). Tests: `test_bootstrap_admin_concurrent_creation_race`, `test_primary_workspace_retries_on_integrity_error` (simulated IntegrityError on first attempt; retry converges to the correct state). | 220501c |
| **4 (low)** — wrong 401 advice + mid-flow abort via the shared API-key helper | The connect flow no longer uses `_create_stardag_api_key` (so `stardag modal stardag-api-key create` keeps its own UX untouched). Its own `_rotate_api_key` raises `RuntimeError` with the failing status/body, and `run_connect` catches it non-fatally: prints the error plus "Modal DAG execution is not wired up yet - re-run `stardag self-host connect` to retry (the rest of the setup is completed below)", then still writes the local SDK registry/profile/caches; the summary panel omits the API-key line. Test: `test_connect_api_key_creation_failure_is_not_fatal` (500 on create → no secret push, profile + registry still written). | 497a7fc |

Verification: `app/stardag-api` 361 passed / 18 skipped; `lib/stardag` `tests/test_selfhost` 53 passed (full collection); pre-commit (ruff, ruff-format, pyright both packages, prettier) clean on all touched files.

https://claude.ai/code/session_01ARrhAGii9mMcpyVNTbSMEf

andhus:
## Dry-run fix: client/image Python version matching (bf1e963)

A from-scratch dry run of `self-host up --from-source` under Python 3.14 hit Modal's `InvalidError: The 'migrate' Function (using serialized=True) was defined with Python 3.14, but its Image has 3.12` — the serialized closures are pickled by the *client* interpreter, but `build_server_app` hardcoded the image to 3.12.

Fixed on both paths:

- **From-source**: the image's `add_python` now defaults to the running interpreter's `major.minor` (`client_python_version()`, override param kept), with a clear error below stardag-api's minimum (>= 3.10).
- **Prebuilt**: new `PREBUILT_IMAGE_PYTHON = (3, 12)` constant (next to `DEFAULT_SERVER_VERSION`); `up`/`upgrade` fail fast *before any provisioning* when the client interpreter differs, printing the remedy (`uvx --python 3.12 --from 'stardag[selfhost]' stardag self-host up`, or `--from-source`).
- **Docs**: quickstart/upgrade/OIDC/connect commands now use `uvx --python 3.12 ...` with a one-line why; troubleshooting entry for the `InvalidError`; DEV_README's "Releasing the Server" documents the Dockerfile ↔ `PREBUILT_IMAGE_PYTHON` coupling (also guarded by a unit test against `app/server.Dockerfile`).
- **Tests**: derived/overridden `add_python`, version-check pass/fail, too-old interpreter error, Dockerfile-constant guard — `tests/test_selfhost` now 60 passed; pre-commit clean.

https://claude.ai/code/session_01ARrhAGii9mMcpyVNTbSMEf

andhus:
## Polish pass (0086c2e)

Three follow-up refinements:

1. **Fixed the doubled default URL.** The Modal app name and its dedicated Modal environment were both `stardag-server`, producing `https://<ws>-stardag-server--stardag-server.modal.run`. Now `DEFAULT_APP_NAME = "server"` and `DEFAULT_SERVER_MODAL_ENV = "stardag-host"` → `https://<ws>-stardag-host--server.modal.run`. Secrets/Dict derive from the app name (`server-config`/`server-jwt`/`server-meta`, env-namespaced). Added an upgrade-section docs note: a deployment created with a non-default `--name`/`--server-modal-env` (e.g. an older one named `stardag-server` in the default Modal env) must be upgraded by passing the same flags (`--name stardag-server --server-modal-env ''`).

2. **Better default target-root URI:** `modalvol://stardag-targets/<workspace-slug>` (was `modalvol://stardag/...`) via a new `default_target_root_uri()` helper. The `<workspace-slug>` segment is deliberately kept — multiple Stardag workspaces can execute in one Modal environment with deterministic output paths, so the slug namespaces them and prevents collisions in the shared volume.

3. **Summary panel next-step link:** "Deploy a DAG app to Modal" now links `https://docs.stardag.com/how-to/integrate-modal/` directly.

Docs, skill bundle, CHANGELOG, and tests updated; added tests for the app/env default divergence and the target-root helper. `lib/stardag` `tests/test_selfhost` 62 passed; `mkdocs build` clean; pre-commit clean.

https://claude.ai/code/session_01ARrhAGii9mMcpyVNTbSMEf

andhus:
Finalized the default target-root format (`171aa00`).

The default target root is now a **dedicated Modal volume per (workspace, environment)**, with the target-root name as the path:

```
modalvol://stardag-targets-<workspace-slug>-<environment-slug>/default
```

(previously `modalvol://stardag-targets/<workspace-slug>`).

Rationale: the volume is a single atomically deletable unit on teardown (`modal volume delete stardag-targets-<ws>-<env>`) rather than leaving orphaned subpaths in a shared volume; `stardag-targets-` stays a globbable namespace; and using the root name as the path lets multiple modalvol roots on one environment (`default`, `scratch`) coexist while non-modal roots are unaffected.

A new `_targets_volume_name` helper enforces Modal's **64-char object-name limit** (charset `[a-zA-Z0-9-_.]`): when the composed name overflows it is truncated and given a short deterministic hash of the (workspace, environment) pair, so the result stays ≤64 chars and distinct pairs never collide. Covered by a new length-guard unit test alongside the updated URI tests.

https://claude.ai/code/session_01ARrhAGii9mMcpyVNTbSMEf

Issue #208, opened 2026-08-07
Reactive scheduling: correctness gaps, operational blind spots, and a missing task-registration story


andhus:
# Reactive scheduling: correctness gaps, operational blind spots, and a missing task-registration story

Consolidated findings from running reactive Modal builds against a
long-lived environment. Most of this is not "reactive scheduling is
broken" — the core loop works. The problems cluster in three places:

1. a handful of genuine correctness bugs where a build fails or wedges
   for reasons that have nothing to do with the user's DAG,
2. an operational blind spot: nothing terminates abandoned builds, and
   there is no CLI/API surface to find or clean them up,
3. a missing feature: reactive scheduling needs task classes to be
   *registered* in the scheduler process, and there is currently no
   supported way for a user to declare them.

Filed as one issue for coherence — (3) is the largest piece of work and
several of the bugs are only fully fixable alongside it. Happy to split
into separate issues per section if preferred.

---

# Part A — Bugs

## A1. A blocker in another build is invisible to the frontier liveness check, so the build fails spuriously

**Severity: high.** This kills otherwise-valid reactive builds within
seconds of triggering, with an error that points nowhere near the cause.

**Symptom.** A reactive build fails almost immediately with:

```
No runnable or running tasks left but roots are not complete
(status counts: {'completed': 2, 'suspended': 1})
```

**Mechanism.** `Task.latest_status` is denormalised on a row unique per
`(environment_id, task_id)` — it is *environment-global*, not per-build.
`TaskDependency` edges are global too. `get_build_frontier` then mixes
the two scopes inconsistently:

- `has_incomplete_upstream` joins `Task` **globally**, so an upstream
  left `RUNNING` by *any* build in the environment correctly blocks a
  downstream task, removing it from `actionable`.
- `running` and `status_counts` are scoped to
  `Task.id.in_(build_task_ids)`, i.e. only tasks with events for *this*
  build.

A task that is running in another build is therefore fully effective as
a blocker but contributes nothing to this build's `running` count. In
`_handle_terminal`, `not frontier.actionable and running == 0` is read as
"the build cannot progress" and the build is failed.

This is easy to hit whenever DAGs overlap across builds, which is the
normal case for a shared environment — and especially when the blocking
task is a *dynamic* dependency registered under the earlier build, so it
is not in the new build's task set at all.

`TickConfig.stale_running_no_ref_seconds` cannot rescue this: it is
evaluated inside `_resolve_running`, which is only reached for tasks the
[...]

PR #209, opened 2026-08-07
fix(reactive): scope the watchdog sweep server-side and fail trigger cleanly (#208 A3/A4)

andhus:
Two independent reactive-scheduling fixes from #208 — sections **A3** and
**A4**. Both are small, and both matter more the longer an environment
lives. `lib/` only.

## A3 — the watchdog ignores the server-side filters that exist for it

`build_list_running` paged `GET /builds` **unfiltered** and filtered on
the derived status client-side, even though the API has supported
`status=` and `reactive_app_name=` since the reactive-metadata release —
`reactive_app_name` exists precisely because "RUNNING reactive builds
owned by app X" is the watchdog's actual question.

One correction to the issue's phrasing, worth stating because it changes
the cost model: `_run_watchdog_sweep` does not itself issue a per-build
`GET /builds/{id}`. It calls the tick **locally, in-process, sequentially**
with `linger_seconds=0`, and the tick's own pre-lease gate
(`registry.build_get`, returning `{"outcome": "not_reactive"}`) is what
discovers most builds aren't reactive. So an irrelevant RUNNING build cost
a whole tick invocation every sweep period, not just a listing row.

The cap is the sharper problem. Builds sort by `last_active_at desc`, so
once an environment holds more stale RUNNING builds than `sweep_limit`
(nothing terminates abandoned builds today — #208 B3), the sweep can stop
reaching genuine reactive builds **entirely and silently**, disabling the
safety net exactly when it's needed.

Changes:

- `RegistryABC.build_list_running` / `build_list_running_aio` /
  `APIRegistry.build_list_running` take an optional `reactive_app_name`,
  and the API implementation now sends `status=running` unconditionally
  plus `reactive_app_name` when given.
- `_run_watchdog_sweep` takes `reactive_app_name` and the deployed
  `tick_watchdog` passes this app's name.
- The tick's own gate is **untouched** — it's still needed for
  `reactive_tick_kwargs` and app-ownership forwarding, and it stays the
  correctness backstop. It's simply now paid only for builds that survived
  the server-side filter.
- The truncation warning distinguishes `N+ reactive builds owned by <app>`
  from `N+ running builds` and says what to do about it.

**Graceful degradation.** A server predating the filters ignores unknown
query params rather than rejecting them, so it answers with the unfiltered
listing. The client-side status re-check is deliberately kept: it bounds
that degradation to "wider than asked for" instead of "the watchdog ticks
terminal builds". A wider listing is otherwise harmless — a tick no-ops on
a non-reactive build. Custom `RegistryABC` implementations written before
the kwarg are detected by signature inspection (the existing `accepts_*`
pattern in `registry/_base.py`) and called the old way.

**Behaviour change a reviewer should weigh.** Scoping drops the incidental
cross-app coverage a sweep used to provide, where app A's watchdog ticked
app B's builds and the tick forwarded to B. That coverage was accidental
and competed for the same cap; the owner app's own watchdog is the
supported mechanism, and forwarding still handles wake-ups from a previous
owner's still-running workers. The one case that regresses is a build
owned by an app deployed **without** `watchdog_period_minutes` — which
`build_trigger` already warns about at trigger time. Documented in both
`integrate-modal.md` and `build-execution.md`.

## A4 — a failed reactive trigger left an orphan RUNNING build forever

`_trigger_reactive` mints the build first, then discovers, then writes the
task store, then writes the reactive marker, then spawns. Build status is
derived from events, so a failure anywhere in that window left a build
`RUNNING` forever with no orchestrator that could ever emit a terminal
event — and, because the marker is written **last**, not even attributable
to an app: invisible to its own app's now-scoped sweep and pure overhead
for every other one.

All post-`build_start` work is wrapped so **any** exception emits a
terminal `BUILD_FAILED` naming the stage that failed
(`"Reactive trigger failed during task store write: PermissionError: …"`)
before the original exception propagates. If emitting that event also
fails, it's logged and the original error still propagates — the root
cause is never masked.

Two deliberate choices:

- **Ordering left alone.** Writing the reactive marker earlier would make
  a wedged build discoverable, but it would also expose a
  partially-discovered DAG to a tick, which treats the registry frontier
  as ground truth. Preferring the terminal-event wrapper also covers
  discovery, which reordering the task-store write alone would not.
- **The spawn is outside the wrapper.** By then the durable state is
  complete and the build carries the reactive marker, so a failed spawn is
  recovered by the next watchdog sweep or a re-trigger. Failing the build
  there would take it out of the watchdog's RUNNING listing and remove
  that recovery path.

`except BaseException` rather than `except Exception`: a Ctrl-C during
discovery is one of the likelier ways to abandon a trigger and wedges the
build identically. The original exception is always re-raised.

#208 B1 will later remove the target-root write from this path for most
users; A4 is worth having independently — discovery can raise too.

## Tests

- `tests/test_registry/test_api_registry.py::TestBuildListRunningFilters` —
  `status=running` always sent; `reactive_app_name` forwarded when given;
  an old server that ignores both filters still yields only RUNNING builds;
  `limit` stops paging mid-page.
- `tests/test_integration/test_modal/test_stardag_app.py::TestWatchdogSweep`
  — the listing is scoped to the app; an implementation predating the
  kwarg degrades to the unscoped call instead of raising; the truncation
  warning names the scope and the remedy.
- `…::TestReactiveTriggerFailureLeavesNoOrphanBuild` — a task-store write
  failure and a discovery failure each emit `BUILD_FAILED` with the right
  stage; a secondary `build_fail` error never masks the original; a spawn
  failure leaves the build marked and recoverable rather than failed.

`uv run pytest tests -m "not modal_live" -q` → **956 passed, 29
deselected**. No live/integration tests were run.

`pre-commit run --files <changed>` → `prettier`, `ruff`, `ruff-format`,
`pyright-stardag` all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,39 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
 ## [Unreleased]
 
+### Fixed
+
+- **The reactive watchdog now asks the registry only for the RUNNING builds
+  its own app owns** (`GET /builds?status=running&reactive_app_name=...`,
+  filters the API has supported since the reactive-metadata release but the
+  SDK never used). Previously each sweep paged the whole build listing,
+  filtered the derived status client-side, and then spent a full `tick`
+  invocation on every RUNNING build in the environment just to discover
+  most were not reactive. Worse, the sweep's per-period cap was consumed by
+  those irrelevant builds: an environment holding more stale RUNNING builds
+  than the cap could stop reaching genuine reactive builds entirely, and
+  silently — disabling the safety net exactly when it was needed. The
+  truncation warning now says how many _reactive_ builds owned by _which
+  app_ it truncated, and what to do about it. `build_list_running` gained an
+  optional `reactive_app_name` argument; the client-side status re-check is
+  retained so a server predating the filters degrades to a wider listing
+  rather than to ticking terminal builds. Note that scoping removes the
+  incidental cross-app coverage a sweep used to provide: a build owned by
+  an app deployed without a watchdog is no longer swept by another app's
+  watchdog. ([#208](https://github.com/stardag-dev/stardag/issues/208) A3)
+- **A reactive trigger that fails part-way no longer leaves a build stuck
+  in RUNNING forever.** `build_trigger(reactive=True)` mints the build
+  before running discovery and persisting the task store, and a build's
+  status is derived from its events — so a failure in between (most often a
+  target-root permission or storage error on the task-store write) left a
+  RUNNING build that nothing would ever terminate and that carried no
+  reactive owner, so it was invisible to its own app's sweep and pure
+  overhead for every other one. All post-mint trigger work is now wrapped:
+  any failure emits a terminal `BUILD_FAILED` naming the stage that failed
+  before the original exception propagates (a failure to record that event
+  is logged, never allowed to mask the root cause).
+  ([#208](https://github.com/stardag-dev/stardag/issues/208) A4)
+
 ## [0.17.0] — 2026-08-06
 
 ### SDK
--- a/docs/docs/concepts/build-execution.md
+++ b/docs/docs/concepts/build-execution.md
@@ -145,10 +145,12 @@ DAG-defining code) spawns work.
 
 Each reactive build is **owned by the app that triggered it** (the owning
 app name is stored in the registry with the build's reactive metadata):
-ticks from any other deployed app in the environment — typically another
-app's watchdog sweep — forward the wake-up to the owner's tick instead of
-driving the build with the wrong code. Ownership moves only by an explicit
-re-trigger from the new app, which also updates the tick config.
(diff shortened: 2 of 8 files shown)


andhus:
Copilot review addressed — both findings were real, and the second one was sharper than it looked.

**`build_list_running_aio` forwarding.** Correct: the signature check only ran when `reactive_app_name` was non-None, so an *unscoped* call still passed `None` as a second positional argument and raised `TypeError` on a pre-kwarg implementation — on precisely the call that asked for no scoping. Now forwards only when there is something to forward, and by keyword, since an implementation may make the parameter keyword-only. Three tests added for the delegate.

**Truncation warning.** Also correct, and the fix exposed a second bug: the remedy was chosen from whether a scope was *asked* for, not whether the listing was actually scoped. A scoping request an older registry cannot honour still yields a listing capped by RUNNING builds of every kind. Both now key on the same `scoped` flag, with a test.

Note for readers of the stack: this compatibility shim is deliberately **removed** two PRs later in #217, where `build_list_running` is re-expressed in terms of `build_list` and the scoping becomes server-side for every registry.

PR #210, opened 2026-08-07
fix(api): surface externally-blocked tasks in the build frontier; allow retry from suspended (#208 A1/A2)

andhus:
Server side of **A1** and **A2** from #208. Registry API only — no SDK or
UI changes. Both changes are purely additive and backward compatible:
existing frontier fields keep their exact semantics and values, and a
client that ignores the new fields behaves exactly as before.

## A1 — scope-consistent liveness in the build frontier

`GET /builds/{id}/frontier` mixes two scopes:

- `has_incomplete_upstream` joins `Task` **globally** — task rows are
  unique on `(environment_id, task_id)` with denormalised `latest_status*`,
  and `TaskDependency` edges are global too — so an upstream left
  non-COMPLETED by *any* build removes a downstream task from `actionable`.
- `running` and `status_counts` are scoped to `Task.id.in_(build_task_ids)`,
  i.e. only tasks this build has events for.

So a task running in another build is fully effective as a blocker while
contributing nothing to this build's `running` count. A scheduler reads
"nothing actionable and `running == 0`" as "this build cannot progress"
and fails it, with an error naming only status counts. The blocker need
not even be in this build's task set — a dynamic dependency registered
under an earlier build is the common case.

This PR does not change the liveness rule; it gives the client the
information the rule was missing, as an explicit `blocked_by_external`
list on the frontier response.

### Response shape (new fields)

```jsonc
{
  // ... all existing BuildFrontierResponse fields, unchanged ...
  "blocked_by_external": [
    {
      "task_id": "…",                     // string  — blocked task, always a member of this build
      "blocking_task_id": "…",            // string
      "blocking_task_namespace": "…",     // string  ("" for the default namespace)
      "blocking_task_name": "…",          // string
      "blocking_status": "running",       // TaskStatus enum value; never "completed"
      "blocking_status_at": "2026-01-01T00:00:00Z", // string|null (ISO-8601 datetime)
      "blocking_status_build_id": "…",    // string|null (UUID) — the build that owns the status
      "blocking_in_build": false          // bool
    }
  ],
  "blocked_by_external_truncated": false  // bool
}
```

Both fields default to `[]` / `false`, so nothing changes for builds with
no external blockers (the overwhelming majority).

Semantics of an entry: `task_id` is a task of this build with global
status PENDING / SUSPENDED / RUNNING, held back by `blocking_task_id`,
which is not COMPLETED and whose current status was **not** produced by
this build (`blocking_status_build_id` is not this build's id, null-safe —
a null, i.e. a row predating status denormalisation, also counts as "not
this build's doing").

`blocking_in_build` says whether the blocker is also in this build's task
set. `false` is the pathological case behind the spurious failures: this
build will never schedule the blocker, so it can only wait for whoever
owns it. `true` still blocks, but the blocker also appears in this build's
own `actionable` / `running`.

### Design notes

- **Name/namespace of the blocker are included**; of the blocked task they
  are not. The blocked task was registered by the caller and every other
  frontier field is `task_id`-keyed, whereas the blocker may be entirely
  unknown to the caller — a diagnostic that only says "blocked by
  `9f3a…`" is not actionable. Both come from the same join, so there is no
  extra round-trip either way.
- **Capped at 50 entries, with `blocked_by_external_truncated`** rather
  than a silent cut. A wide DAG stalled behind another build can produce
  one entry per blocked edge; the list is a diagnostic ("you are waiting,
  and here is on what"), so a bounded, deterministically-ordered sample
  plus an explicit flag carries the same signal at a fixed payload size. A
  total count was considered and dropped: it needs a second aggregate over
  the same predicate, and the flag already distinguishes "this is all of
  them" from "there are more".
- **Ordering** is registration order (blocked task, then blocker), matching
  `actionable`, so the truncated prefix is stable across the polls of a
  single tick.
- **Reported separately rather than folded into `running`.** `running` must
  keep meaning "RUNNING tasks of *this* build" — it is the list a tick
  cancels through.
- **A PENDING upstream this build merely referenced also qualifies**, since
  its PENDING was recorded by an earlier build. That is accurate (this
  build did not put it there) but is not a stall — such a blocker is
  in-build and actionable. Consumers can discriminate on
  `blocking_status` + `blocking_in_build`.

### Cost

One additional query per frontier read. It is a flat
`task_dependencies ⋈ tasks ⋈ tasks` join restricted to this build's task
set (`ix_task_dep_downstream`, then the tasks PK) with a `LIMIT` —
essentially the same join `actionable` already performs per row as a
correlated `EXISTS`, done once and non-correlated. Resolving blockers per
blocked task instead would be N+1 on an endpoint polled on every tick and
every linger poll, which is why it is one statement. The endpoint already
issues five queries; this makes six, and the new one touches no rows the
others do not.

### Also fixed here: `latest_status_at` was always null

`FrontierTaskRef` declares `latest_status_at` and documents it as the
input to scheduler staleness bounds ("RUNNING too long with no executor
ref"), but the frontier never populated it — the field defaults to `None`,
so it silently serialised as `null` on every ref and those guards were
dead. Now populated from `Task.latest_status_at` for `actionable`,
`running` and `roots`. Same area, same class of bug (declared signal,
never sent), so it rides along.

## A2 — `suspended` is recoverable

A task left `SUSPENDED` — suspended for dynamic dependencies, then
abandoned because the orchestrator died or the build was cancelled — was
permanently unschedulable. `TASK_RETRIED` reset only FAILED / CANCELLED /
SKIPPED, so the retry endpoint was a no-op on it. The only escape was to
cancel the task purely to reach a status that *was* retryable and then
retry — undocumented and non-obvious.

`suspended` now resets to `PENDING`. A suspended task is not executing —
the suspension means the execution yielded and returned — so resetting it
cannot orphan a live execution, and re-running from scratch is exactly
what a retry expresses. The retry also clears the executor ref of the run
that suspended, which is what stops a scheduler from re-attaching to an
execution that will never resume.

**`running` remains non-retryable, deliberately.** It holds a live
execution claim; releasing that claim is cancellation, not retry, and
conflating them risks a second concurrent execution of the same task. This
is stated in the endpoint docstring (and therefore the OpenAPI
description) and beside the constant.

The retryable set lived in three hard-coded tuples — the denormalised
`apply_event_to_task` path and the two per-build event replays, which must
agree — now hoisted into one `_RETRYABLE_STATUSES` so they cannot drift.

Note this is the *server* half. The SDK's own `_RETRYABLE_STATUSES`, used
by `discover_and_register_aio(..., retry_failed=True)` to decide which
tasks to retry, is unchanged and lands separately.

## Tests

Added to `app/stardag-api/tests/test_builds.py`:

- `test_frontier_reports_blocker_owned_by_another_build` — the end-to-end
  A1 repro: build A registers `up → down` and starts `up`; build B
  references only `down`. Build B's frontier has empty `actionable`, empty
  `running` and `status_counts == {"pending": 1}` (the shape that used to
  read as terminal) and now reports the blocker with build A's id and
  `blocking_in_build: false`. Build A's own frontier reports no external
  blockers. Completing `up` clears the list and makes `down` actionable.
- `test_frontier_reports_blocker_owned_by_another_build_postgres` — the
  same scenario against Postgres via the `pg_client` fixture, because the
  "not this build's status" predicate is a null-safe inequality that the
  two dialects render differently (`IS DISTINCT FROM` vs `IS NOT`).
- `test_frontier_external_blocker_reports_in_build_membership` — a blocker
  the build does know about is still reported, flagged `blocking_in_build:
  true`.
- `test_frontier_external_blockers_are_capped_and_flagged` — cap and
  truncation flag, with a stable prefix.
- `test_frontier_refs_carry_status_timestamp` — regression test for the
  always-null `latest_status_at`.
- `test_retry_task_resets_suspended_to_pending` — suspended → pending
  through the endpoint, executor fields cleared, and the per-build event
  replay agreeing with the denormalised global status.

Existing `test_retry_task_never_downgrades_completed_or_running` still
covers the completed/running no-ops.

No migration: query-only, no model change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,12 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
   was declared and documented as the input to scheduler staleness bounds
   but never populated, so it always serialised as `null` and those guards
   silently did nothing.
+- **`POST /builds/{id}/tasks/{task_id}/retry` accepts `suspended`.** A task
+  suspended for dynamic dependencies and then abandoned (orchestrator died,
+  build cancelled) was permanently unschedulable — the only escape was an
+  undocumented cancel-then-retry. `running` remains non-retryable on
+  purpose: it holds a live execution claim, and releasing that is
+  cancellation, not retry.
 
 ### SDK
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,25 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
 ## [Unreleased]
 
+### Registry API
+
+- **`GET /builds/{id}/frontier` now reports why a build is waiting on work
+  it does not own** ([#208](https://github.com/stardag-dev/stardag/issues/208)).
+  Dependency gating is environment-global (task rows and edges are shared
+  across builds) while `running`/`status_counts` cover only the tasks a
+  build has events for, so an upstream left non-COMPLETED by another build
+  could gate a build's tasks while contributing nothing it could see — a
+  scheduler then read "nothing actionable, nothing running" as "cannot
+  progress" and failed the build. The new `blocked_by_external` list pairs
+  each such blocked task with its blocker (identity, status, the owning
+  build id and status timestamp, and whether the blocker is in this build's
+  task set), capped with an explicit `blocked_by_external_truncated` flag.
+  Purely additive: every existing field keeps its exact semantics.
+- **Frontier task refs now actually carry `latest_status_at`.** The field
+  was declared and documented as the input to scheduler staleness bounds
+  but never populated, so it always serialised as `null` and those guards
+  silently did nothing.
+
 ### SDK
 
 - **Reactive builds now discover the DAG inside Modal, not on the machine
--- a/app/stardag-api/src/stardag_api/routes/builds.py
+++ b/app/stardag-api/src/stardag_api/routes/builds.py
@@ -2524,14 +2524,25 @@ async def retry_task(
     auth: Annotated[SdkAuth, Depends(require_sdk_auth)],
     commit_hash: str | None = None,
 ):
-    """Reset a failed/cancelled/skipped task to pending (retry).
+    """Reset a failed/cancelled/skipped/suspended task to pending (retry).
 
-    Emits TASK_RETRIED; status derivation flips only terminal-but-
-    retryable statuses back to PENDING — completed and running tasks are
-    unaffected (the event is still recorded, making concurrent
-    trigger/retry races benign). Used by reactive triggers so a
-    re-triggered failed build (or a new build referencing a previously
-    failed task) becomes schedulable again.
+    Emits TASK_RETRIED; status derivation flips only retryable statuses
+    back to PENDING. Used by reactive triggers so a re-triggered failed
+    build (or a new build referencing a previously failed task) becomes
+    schedulable again.
+
+    **Suspended tasks are retryable.** A task suspended for dynamic
+    dependencies is not executing — the execution yielded and returned —
+    so a task whose orchestrator then died has no path forward except
+    running again from scratch, which is what a retry means. Without this
+    it would be permanently unschedulable.
+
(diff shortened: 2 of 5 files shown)


andhus:
Copilot raised no inline findings on this PR — review body only. Nothing to resolve.

PR #211, opened 2026-08-07
feat(api): persist reactive scheduler tick summaries per build (#208 B6)

andhus:
Implements **B6** of #208 ("Tick outcomes are not persisted"), server side only.

A reactive build is driven by many short-lived scheduler ticks, each in its own container. Each tick's `TickSummary` — the scheduler's own account of what it did, and of why it did nothing — reached nothing but that container's log, so reconstructing why a build stalled meant correlating logs across dozens of containers. This persists the last N summaries against the build and exposes them.

Scope is `app/stardag-api/` only. Nothing writes to these endpoints yet: SDK-side reporting and the UI panel (B5) land separately, and both depend on the shapes below **verbatim**.

### `POST /api/v1/builds/{build_id}/tick-summaries`

Auth: SDK (API key or internal JWT), scoped to the build's environment.

Request body — an **open object**. `outcome` is the only required key (`maxLength: 32`); everything else is accepted as-is and stored unchanged, including keys this server has never heard of. Today's `TickSummary` serialises to:

```json
{
  "outcome": "lingered_out",
  "terminal_status": null,
  "spawned": 0,
  "self_healed": 0,
  "failed_recorded": 0,
  "cancelled_refs": 0,
  "iterations": 1,
  "limit_denied": 0,
  "claim_denied": 0,
  "skipped": 0
}
```

`201` response body:

```json
{
  "id": "0198f0a1-...",
  "build_id": "0198f09c-...",
  "outcome": "lingered_out",
  "summary": {
    "outcome": "lingered_out",
    "terminal_status": null,
    "spawned": 0,
    "self_healed": 0,
    "failed_recorded": 0,
    "cancelled_refs": 0,
    "iterations": 1,
    "limit_denied": 0,
    "claim_denied": 0,
    "skipped": 0
  },
  "created_at": "2026-08-07T20:51:10.251057Z"
}
```

`summary` is the request body verbatim, `outcome` included — the top-level `outcome` is a duplicate of it, promoted for indexing.

Errors: `404` unknown build, `403` build in another environment, `422` missing/over-long `outcome` or a body over 8192 bytes as compact JSON, `429` workspace rate limit.

### `GET /api/v1/builds/{build_id}/tick-summaries?limit=<n>`

`limit` defaults to `20`, `ge=1`, `le=200`. Newest first.

```json
{
  "build_id": "0198f09c-...",
  "summaries": [
    {
      "id": "0198f0a1-...",
      "build_id": "0198f09c-...",
      "outcome": "lingered_out",
      "summary": { "...": "as posted" },
      "created_at": "2026-08-07T20:51:10.251057Z"
    }
  ]
}
```

A build that has never been ticked returns `200` with `"summaries": []`. Errors: `404` unknown build, `403` cross-environment.

## Design notes

**One promoted column plus a blob.** `outcome` is lifted into an indexed `String(32)`: it is the one field with a small, closed-ish vocabulary worth filtering on, and the UI reads it on every row. Everything else lives in a JSON/JSONB blob stored as sent. `TickSummary` is SDK-owned and still growing (per-blocker detail is coming), this table is a write-mostly observability trail rather than a queried entity, and unknown keys must be preserved rather than rejected — so a new field must cost neither a server release nor a migration. `outcome` stays a plain string, not an enum: the Modal wrapper already emits an outcome the core doesn't, and an unrecognised one must round-trip rather than 500.

**Retention pruned on insert**, to the newest N per build (default 50, `STARDAG_API_MAX_TICK_SUMMARIES_PER_BUILD`), inside the insert's transaction. This service has no scheduler to run a background reaper, and the property actually needed is "a build's trail is bounded" — insert-time pruning gives that exactly and immediately. The delete is bounded by construction: steady state removes one row per insert.

**Cheap by contract.** The write sits on a hot path and is best-effort — a caller reports and moves on, and must never fail a tick because this failed. It is one insert plus one bounded delete, nothing more.

**Indexing.** One composite index `(build_id, created_at)` and nothing else. "Newest N for this build" is the only query — for both the read endpoint and the prune — and the composite's leading column also serves plain `build_id` lookups and the FK cascade. Deviates slightly from a separate `build_id` index: the write path pays for every index twice per tick, and a redundant one buys nothing. Direction is left ASC rather than `DESC`; a btree is scanned backwards at no cost, and an expression-based index would make Alembic's autogenerate emit a spurious drop/create on every future run.

**Route module.** New `routes/tick_summaries.py` rather than more of `routes/builds.py` (already 2.8k lines). Same `/builds` prefix and `builds` tag, so it groups with the rest in the OpenAPI schema. It reuses `_get_build_checked` and `_raise_if_limit_exceeded` from `routes/builds.py` rather than copying them — the environment-scoping check in particular should have exactly one implementation.

## Migration

`0807aacc230e` (`20260807_205110_add_build_tick_summaries.py`), autogenerated, revising `792fb8ab3c3c`. New table, no backfill, no changes to existing tables. Verified against a real Postgres 16: `alembic upgrade head && alembic downgrade -1 && alembic upgrade head`, plus `alembic check` reporting no drift afterwards.

## Tests

`app/stardag-api/tests/test_tick_summaries.py` — write/read round trip, unknown-field preservation, retention at the boundary (at N, N+1, and well past), retention scoped per build, size-cap and over-long-`outcome` rejection, `limit` bounds, unknown build → 404, cross-environment write and read → 403, empty trail, and a Postgres round trip via `pg_client` exercising JSONB and the `NOT IN (… LIMIT n)` prune on the production dialect.

```
uv run pytest                                        372 passed, 19 skipped
STARDAG_API_TEST_DATABASE_URL=… uv run pytest        390 passed,  1 skipped
pre-commit run --files <changed files>               all hooks passed (ruff, ruff-format, pyright-stardag-api, prettier)
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,18 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
 ### Registry API
 
+- **Reactive scheduler tick summaries are now persisted per build**
+  (`POST`/`GET /builds/{build_id}/tick-summaries`). A reactive build is
+  driven by many short-lived scheduler ticks, each in its own container,
+  and each tick's `TickSummary` — the scheduler's own account of what it
+  did and why it did nothing — previously reached nothing but that
+  container's log. Reconstructing why a build stalled meant correlating
+  logs across dozens of containers; it is now a single request. The
+  summary is stored verbatim in a JSONB blob (with `outcome` promoted to
+  an indexed column), so the SDK can add fields without a server release
+  or a migration. Retention is bounded per build (newest 50 by default,
+  `STARDAG_API_MAX_TICK_SUMMARIES_PER_BUILD`), pruned on insert.
+  Additive: nothing writes to these endpoints yet.
 - **`GET /builds/{id}/frontier` now reports why a build is waiting on work
   it does not own** ([#208](https://github.com/stardag-dev/stardag/issues/208)).
   Dependency gating is environment-global (task rows and edges are shared
--- a/app/stardag-api/migrations/versions/20260807_205110_add_build_tick_summaries.py
+++ b/app/stardag-api/migrations/versions/20260807_205110_add_build_tick_summaries.py
@@ -1,0 +1,60 @@
+"""add build tick summaries
+
+Revision ID: 0807aacc230e
+Revises: 792fb8ab3c3c
+Create Date: 2026-08-07 20:51:10.251057
+
+"""
+
+from typing import Sequence, Union
+
+from alembic import op
+import sqlalchemy as sa
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision: str = "0807aacc230e"
+down_revision: Union[str, Sequence[str], None] = "792fb8ab3c3c"
+branch_labels: Union[str, Sequence[str], None] = None
+depends_on: Union[str, Sequence[str], None] = None
+
+
+def upgrade() -> None:
+    """Upgrade schema."""
+    # Per-build trail of reactive scheduler tick outcomes. ``outcome`` is
+    # promoted out of the payload (small closed-ish vocabulary, worth
+    # filtering on); ``summary`` keeps the SDK-owned dict verbatim so new
+    # fields need no migration. New table, no backfill.
+    op.create_table(
+        "build_tick_summaries",
+        sa.Column("id", sa.Uuid(), nullable=False),
+        sa.Column("build_id", sa.Uuid(), nullable=False),
+        sa.Column("outcome", sa.String(length=32), nullable=False),
+        sa.Column(
+            "summary",
+            sa.JSON().with_variant(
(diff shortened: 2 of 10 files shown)


andhus:
Copilot review triaged.

**Retention pruning race.** The analysis is correct: two concurrent inserts for the same build can each compute `retained_ids` without seeing the other, so the row count can transiently exceed `max_tick_summaries_per_build`.

My inclination is **not** to serialise on the build row for this. Taking a `SELECT … FOR UPDATE` on every tick-summary write puts a lock on the hot scheduler path to defend a *retention* bound whose violation is cosmetic and self-correcting on the next insert. Worth a second opinion before I close it either way.

andhus:
**Now resolved** (my earlier comment said not-yet-implemented).

Kept the design — no row lock on the tick path — and corrected the comment, which claimed a build "never observes an over-long trail". It is an *eventual* bound: concurrent inserts for the same build can briefly exceed the cap, and the next insert prunes it. The window is already narrow because the scheduler lease single-flights ticks per build.

PR #212, opened 2026-08-07
feat(reactive): declare task modules so scheduler ticks can reconstruct tasks without pickles (#208 B1)

andhus:
Implements **#208 B1** — task registration for reactive scheduling — in three
reviewable steps plus docs.

## The problem

A reactive scheduler tick is a fresh, short-lived process. It learns *which*
tasks are actionable from the registry, but to spawn a worker it needs the
actual task *object*, and it has two ways to get one:

| path | fails when |
| --- | --- |
| pickle from the `BuildTaskStore` | the app was redeployed (pickles are same-deployment only), or the target root isn't writable |
| `task_from_registry_data` | the module defining the task class was never imported in the reconstructing process |

The second failure mode is easy to miss and hard to diagnose. A pickle embeds
`module.QualName` and `pickle.loads` **self-imports**, so it resolves its class
regardless of what the container imported. Polymorphic JSON carries only
`__namespace` / `__name`; `get_class()` is a plain dict lookup that raises
`KeyError` and **never attempts an import** — and since the default namespace is
`""`, the payload generally contains no module locator at all. Classes register
at *class-definition* time, so the only way to make one resolvable is to import
its defining module.

Meanwhile the deployed tick is defined inside stardag and drags in user modules
only incidentally (whatever the app's selector callables transitively import).
That's arbitrary, and usually doesn't cover a DAG's task classes — DAGs are
typically assembled in scripts and entrypoints, not in the module defining the
app. So today the pickle path is load-bearing.

### (i) Declaration and deploy-time expansion — `449d8ca`

New executor-agnostic module `stardag/build/_task_modules.py` holding the whole
concept, plus `StardagApp(task_modules=[...])`.

- **Pattern grammar**: an exact module (`"my_pkg.tasks.ingest"`) or a package
  with a trailing recursive wildcard (`"my_pkg.tasks.*"`). A `*` anywhere but
  the final component, an empty/whitespace-padded pattern, a bare `*`, or a
  component that isn't an identifier is a `TaskModulesError` raised eagerly from
  `StardagApp.__init__` — a typo must never degrade into a silent no-match,
  because the symptom would surface hours later as a tick that can't rebuild a
  task.
- **Default** (`None`): infer `"<root package of the module defining the app>.*"`
  from the calling frame. A module with no package (`__main__`, a loose deploy
  script) has no importable name in a container, so inference **warns and opts
  out** rather than baking a list that would fail to import everywhere. `[]`
  opts out silently.
- **Expansion imports nothing it doesn't have to.** `pkgutil.walk_packages` is
  deliberately avoided — it imports each package to reach its `__path__`. Only
  the root package of a `"pkg.*"` pattern is imported (unavoidable); the rest is
  a filesystem walk with `pkgutil.iter_modules`, guarded against symlink loops.
  `__main__` submodules are skipped (side-effectful entrypoints);
  `_`-prefixed modules are **not** (a user's `_tasks.py` is legitimate).
- `finalize()` expands once and bakes the concrete sorted list into the tick
  closure — the same freeze-into-locals pattern already used for `app_name` and
  the selectors — and surfaces it on `FinalizeResult`. In the container the tick
  imports that list before `run_tick_aio`, cached per module list so a container
  serving many ticks pays once. **A module that fails to import warns rather
  than aborting**, and the failures are retained so `_load_task` can append them
  to a rehydration failure ("class X unresolved; note these modules failed to
  import: …"). That annotation is read from the task-module registry rather than
  plumbed through `rehydrate.py`, which stays a pure reconstruction primitive.
- `stardag modal deploy` reports the expansion
  (`task modules: 37 discovered from "my_pkg.*"  ->  128 task classes
  registered`). The class count needs a local import and the deploy environment
  may lack extras the image has, so the check is default-on but **strictly
  warn-only** and can never fail a deploy (`--no-check-task-modules` skips it).
  It lives in the CLI so programmatic `finalize()` callers pay only for name
  expansion.

Also adds `_TypeRegistry.classes()`, the read-only snapshot the class count uses.

### (ii) Trigger-time coverage pre-flight — `c325897`

The trigger is the one place holding both the real DAG and the app, so it's the
only place that can catch a class no tick will be able to reconstruct — before a
container starts. It runs on the **discovered incomplete** set straight after
`discover_and_register_aio`: those are exactly the tasks a tick ever rehydrates,
and reusing discovery's walk avoids a second local DAG traversal. The message
names the class, the app's current patterns, and the narrowest pattern that
would fix it, plus the redeploy requirement.

**Deviation from #208, deliberate: this is a warning, not an error.** #208 frames
it as a hard failure, but an uncovered class isn't broken — it falls back to the
pickle path, which is exactly how every reactive build worked before this
feature. Failing the trigger would break working setups the moment they upgrade,
and "additive" is the premise here. The hard failure is opt-in via
`require_pickle_free=True`. (The middle ground of "error only if the pickle write
failed" is deliberately *not* implemented: it couples error reporting to
storage-failure ordering.)

Dynamically yielded deps don't exist until their parent runs, so the trigger's
pre-flight structurally cannot see them. The deployed worker wrapper therefore
publishes the app's patterns into the process and
`_WorkerLifecycleReporter._register_dynamic_deps` re-runs the check there — once
per class per process, since it runs on every suspending worker invocation.

### (iii) Conditional pickle elision — `f2a988a`

For each task the trigger would persist:

- covered by the patterns **and**
  `task_from_registry_data(task.model_dump(mode="json"), expected_task_id=task.id)`
  round-trips → **write no pickle**;
- otherwise → write the pickle exactly as today.

The self-check is faithful because registration stores exactly
`model_dump(mode="json")` (`_get_task_data_for_registration`), so it's a local
dry run of what the tick will do. It's also *cheaper* than the write it
replaces: pure CPU, versus a per-task target-root existence check plus a
conditional upload. A build whose classes are all covered writes nothing to the
target root, so reactive triggering stops needing write access there — which
also removes the storage-error path that can mint an orphan RUNNING build
(#208 A4). One summary line is logged: how many were pickle-free, how many
pickled, and why.

**What stays pickle-bound, by design** (all of it falls out of the self-check —
no special-casing):

- `AliasTask` payloads — their `loads_type` is pickled bytes, and rehydration
  refuses `__aliased` data because auto-unpickling registry-supplied bytes in a
  scheduler tick is an execution primitive;
- dynamically generated or otherwise non-importable classes;
- anything whose serialization doesn't round-trip to the same task id (notably
  nested task fields with plain task-typed annotations rather than
  `sd.TaskLoads` / `sd.SubClass`).

`require_pickle_free=True` raises naming every task that would have needed a
pickle and why. It's rejected up front alongside `task_modules=[]` (nothing could
ever qualify), and it's enforced **at the trigger only**: a worker registering
dynamic deps applies the same elision but never raises — its task has already
run, and failing the bookkeeping to enforce a storage preference would be
strictly worse than one extra pickle. Dynamic deps do get the elision, so a build
with dynamic dependencies is pickle-free end to end rather than only until the
first `yield`.

Nothing is added to hash-mode serialization — a new top-level key in
`model_dump(mode="json", context={"mode": "hash"})` would silently repartition
every task identity in existence. (The `__module` hint alternative rejected in
the #208 appendix is not implemented.)

### Docs — `f6768e8`

A "Declaring your task modules" section in the Modal how-to, with two caveats
called out explicitly:

- **Task modules become import-hot** — imported in every tick container on every
  cold start, so heavy runtime dependencies belong inside `run()` rather than at
  module scope.
- **Stale-deploy blind spot** — the pre-flight reads the *local* app definition
  and cannot see that the deployed app was built from an older `task_modules`,
  or from a stardag version predating the feature. Since the default *infers*
  patterns, an upgraded SDK triggering against a not-yet-redeployed app can skip
  a pickle the deployed tick can't compensate for. Redeploy before triggering
  (reactive mode already requires the deployed app and SDK to match), or pass
  `task_modules=[]`. Flagged in the CHANGELOG as an upgrade note.

## Scope and compatibility

`lib/` and `docs/` only; nothing under `app/`. Resident (non-reactive) builds
never touch any of this: they hold the real task objects, and workers receive
tasks by value. `task_modules=[]` reproduces the old behaviour byte for byte.
The registry/server needs no knowledge of the feature — `task_data` stays an
opaque blob server-side.

## Test coverage

- `tests/test_build/test_task_modules.py` (new, 52 tests): pattern validation
  including every rejected form; expansion against a **real package tree written
  to disk** and put on `sys.path`, whose modules record their own import to a
  marker file — so the tests can assert that expanding `"pkg.*"` imported the
  root package **and nothing else**, that `__main__` is skipped and `_tasks` is
  not, and that an exact pattern imports nothing at all; import-failure
  retention and the diagnostic note; per-module-list import caching; coverage
  matching (including package-boundary vs. prefix); elision planning, with
  `AliasTask` verified to fail the self-check.
- `tests/test_build/test_reactive.py`: `_load_task`'s diagnostic augmentation,
  with and without recorded import failures, using the existing
  `FakeReactiveRegistry` / `InMemoryTaskStore` / `FakeTickExecutor` fakes.
- `tests/test_integration/test_modal/test_stardag_app.py`: `StardagApp`
  validation and inference (both branches, via controlled frame globals),
  `finalize()` expansion surfaced on `FinalizeResult`, the tick importing exactly
  the baked list (and nothing when opted out), the worker publishing the
  patterns, the pre-flight (warns / doesn't warn / opted out / only-incomplete),
  pickle elision (covered, uncovered, opted out, `AliasTask` DAG),
  `require_pickle_free` raising with every task named, and dynamic-dep elision
  with the once-per-class warning.

`uv run pytest tests -q -m "not modal_live"` → 1026 passed, 29 deselected.
`uv run pyright src tests` → 0 errors. `pre-commit` (ruff, ruff-format,
pyright-stardag, prettier) clean on every changed file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,12 +37,14 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
   about classes the patterns don't cover, naming the pattern to add;
   `require_pickle_free=True` turns that fallback into a hard error.
 
-  Resident (non-reactive) builds are unaffected, and `task_modules=[]`
-  behaves exactly as before. **Upgrade note for reactive users:** because
-  the default _infers_ patterns, a reactive trigger from the upgraded SDK
-  may skip pickles that the currently _deployed_ tick cannot compensate
-  for — redeploy the app before triggering (reactive mode already requires
-  the deployed app and the SDK to match), or pass `task_modules=[]`.
+  Skipping pickles requires declaring `task_modules` explicitly; the
+  inferred default only drives the coverage warning. Upgrading stardag
+  therefore changes nothing on its own — a newer SDK triggering against an
+  app deployed by an older one still writes pickles, because eliding them
+  would depend on a baked-in module list that deployment does not have.
+  Resident (non-reactive) builds are unaffected either way, and
+  `task_modules=[]` behaves exactly as before. **Redeploy the app whenever
+  you change `task_modules`**, before triggering.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,44 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
 ## [Unreleased]
 
+### SDK
+
+- **`StardagApp(task_modules=[...])`: declare the modules whose import
+  registers your task classes, so reactive scheduler ticks can rebuild
+  tasks from registry data instead of pickles.** A tick reconstructs task
+  objects from the registry's stored payload, which resolves a class
+  through the polymorphic registry — populated only as a side effect of
+  importing the defining module. Without a declaration, whatever a tick
+  container happens to import is arbitrary, so the build task store's
+  pickles were load-bearing (and needed target-root write access at
+  trigger time, and were invalidated by every redeploy).
+
+  Patterns are exact modules (`"my_pkg.tasks.ingest"`) or trailing
+  recursive wildcards (`"my_pkg.tasks.*"`); the default infers the root
+  package of the module defining the app, and `[]` opts out. They are
+  expanded to a concrete module list at deploy time (without importing
+  submodules) and baked into the deployed tick, so **adding or moving task
+  classes requires a redeploy**; `stardag modal deploy` reports the
+  expansion (`--no-check-task-modules` skips the warn-only local import
+  check). Task modules are imported in every tick container, so keep heavy
+  runtime dependencies inside `run()` rather than at module scope.
+
+  With the declaration in place, a reactive trigger writes **no pickle**
+  for any task whose class is covered and whose payload round-trips to the
+  same task id — a fully covered build needs no target-root write access
+  at all. Everything else keeps its pickle exactly as before, including
+  `AliasTask` payloads (pickled `loads_type`, never auto-unpickled from
+  registry data by design) and non-importable classes. The trigger warns
+  about classes the patterns don't cover, naming the pattern to add;
+  `require_pickle_free=True` turns that fallback into a hard error.
+
+  Resident (non-reactive) builds are unaffected, and `task_modules=[]`
+  behaves exactly as before. **Upgrade note for reactive users:** because
+  the default _infers_ patterns, a reactive trigger from the upgraded SDK
+  may skip pickles that the currently _deployed_ tick cannot compensate
+  for — redeploy the app before triggering (reactive mode already requires
+  the deployed app and the SDK to match), or pass `task_modules=[]`.
+
 ### Fixed
 
 - **The reactive watchdog now asks the registry only for the RUNNING builds
--- a/docs/docs/how-to/integrate-modal.md
+++ b/docs/docs/how-to/integrate-modal.md
@@ -584,12 +584,24 @@ by default but **warn-only** — your deploy environment may lack extras the
 image has, so a local import failure never fails the deploy. Pass
 `--no-check-task-modules` to skip the check and report names only.
 
-**What you get.** At trigger time, every discovered task whose class is
-covered _and_ whose payload round-trips to the same task id is persisted
-**without a pickle**. A build whose classes are all covered writes nothing
-to the target root, so reactive triggering stops needing target-root write
-access at all. Set `require_pickle_free=True` to turn the fallback into a
-hard error that names every task that would have needed a pickle and why.
+**What you get.** Once you declare `task_modules` explicitly, every
+discovered task whose class is covered _and_ whose payload round-trips to
+the same task id is persisted **without a pickle**. A build whose classes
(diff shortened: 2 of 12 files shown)


andhus:
Copilot review addressed — all three findings were real.

- **Frame reference.** Now scoped in a `try/finally` with an explicit `del`. This runs in long-lived scheduler containers, so leaving frames (and their reference cycles) behind is worth avoiding.
- **Unvalidated patterns.** Fixed, and it mattered more than it reads: a whitespace-padded pattern matches no module while still looking like a declaration at the call site, so it silently returns ticks to the pickle path — the exact failure this feature exists to remove. Patterns are stripped; an empty one raises rather than being stored inert.
- **Leaky test reset.** Fixed — `_reset_import_state_for_tests` now clears the ambient declaration and the one-shot warning set too, which is what made `only_unwarned=True` order-dependent.

Three tests added covering normalisation, refusal, and the full reset.

PR #213, opened 2026-08-07
feat(api): task claim enumeration, cascading cancel, and a stale-build reaper (#208 B2/B3)

andhus:
Server side of **B2** and **B3** from #208. Registry API only — no SDK or
UI changes; the `stardag builds`/`stardag tasks` CLI and the UI columns
that consume this land in separate PRs. Stacked on #210 (base:
`fix/frontier-external-blockers-and-suspended-retry`), so the diff shows
only this delta.

Everything is additive except one changed default-free behaviour: none.
Existing fields keep their exact semantics, existing ordering is unchanged
unless a new filter is passed, and `cascade` defaults off.

---

## B2 — enumerate tasks by status

`GET /tasks` supported `task_name` and `task_namespace` only, and
`TaskResponse` carried no status at all. `latest_status` is
environment-global (task rows are unique per `(environment_id, task_id)`),
so a task left RUNNING by a build whose orchestrator died denies the
execution claim to *every* future build that needs it, indefinitely — and
the only way to find those was to enumerate RUNNING builds and read each
frontier, which misses any claim whose owning build has already gone
terminal.

### `GET /api/v1/tasks`

| param | type | default | notes |
| --- | --- | --- | --- |
| `page` | int ≥ 1 | 1 | unchanged |
| `page_size` | int 1–100 | 20 | unchanged |
| `task_name` | string | – | unchanged |
| `task_namespace` | string | – | unchanged |
| **`status`** | `TaskStatus`, **repeatable** | – | `?status=running&status=suspended` matches either. Unknown value → 422. |
| **`status_older_than`** | ISO-8601 datetime | – | `latest_status_at < status_older_than`. Rows with a null `latest_status_at` never match. |

Ordering: unchanged (`created_at` desc) with no new filter; **oldest claim
first** (`latest_status_at` asc, NULLS LAST, `id` asc tiebreak) as soon as
`status` or `status_older_than` is given.

`TaskResponse` — used by `GET /tasks`, `GET /tasks/{task_id}`,
`POST /builds/{id}/tasks`, `POST /builds/{id}/tasks/bulk` — gains three
fields, all defaulted:

```jsonc
{
  // ... all existing TaskResponse fields, unchanged ...
  "latest_status": "running",                     // TaskStatus | null
  "latest_status_at": "2026-01-01T00:00:00Z",     // string | null (ISO-8601)
  "latest_status_build_id": "…"                   // string | null (UUID)
}
```

`latest_status_build_id` is the claim holder — "running under build Y since
T". #208 B5 calls it "the single most useful piece of information for the
whole class of cross-build blocking problems"; it existed on the model and
on the build frontier but was unavailable anywhere a client could ask about
a task directly.

### Why a timestamp and not a duration

`status_older_than` takes an absolute ISO-8601 cutoff. A duration would
make the filter non-reproducible across pages — the server re-evaluates
`now() - duration` on every request, so rows drift in and out — and
unanswerable from the request alone. Clock skew is the counter-argument,
but skew is seconds under NTP while realistic staleness thresholds are
hours to days, and a CLI converts `--older-than 24h` locally.

The reaper below takes the *opposite* choice (`idle_for_seconds`)
deliberately: its threshold is a recurring policy, and a fixed timestamp
would be stale on the second invocation of a timer. The unit is in the
name, as required.

### Index

New composite `ix_tasks_environment_status` on
`(environment_id, latest_status, latest_status_at)` — migration
`11716b1f4c1a`, `alembic revision --autogenerate`.

Warranted: the existing single-column `latest_status` index spans every
environment, so "RUNNING" selects every workspace's running tasks before
the environment filter narrows them; the environment-keyed composites
don't mention status at all. The status distribution is heavily skewed —
COMPLETED dominates a mature environment while RUNNING is a handful of
rows — which is exactly where a composite pays. `latest_status_at` is the
third column so the range predicate and the single-status ORDER BY resolve
from the index; a multi-value status filter still sorts, but over a tiny
row set.

---

### 1. `cascade` on `POST /api/v1/builds/{build_id}/cancel`

| param | type | default |
| --- | --- | --- |
| `triggered_by_user_id` | string | – (unchanged) |
| `commit_hash` | string | – (unchanged) |
| **`cascade`** | bool | **`false`** |

Response is `BuildCancelResponse` — a **superset** of `BuildResponse`, so
existing clients are unaffected:

```jsonc
{
  // ... all existing BuildResponse fields ...
  "last_active_at": "2026-01-01T00:00:00Z",   // string|null — see below
  "last_activity_at": "2026-01-02T00:00:00Z", // string|null — see below
  "cascaded_task_ids": ["…"],                 // string[]  — [] unless cascade=true
  "cascaded_task_count": 1                    // int
}
```

With `cascade=true`, TASK_CANCELLED is emitted in the same transaction for
the build's tasks that are RUNNING or SUSPENDED **and whose current status
this build produced** (`latest_status_build_id == build_id`). Two
deliberate restrictions, both about not breaking somebody else's build:

- **Never another build's running task.** The server cannot stop a live
  worker — it only rewrites the registry's view — so declaring another
  build's running task dead is the worst available outcome. Releasing that
  claim is *that* build's cancel.
- **PENDING is not cascaded.** It holds no claim and no slot, and
  ownership scoping cannot protect it: a task this build registered may be
  referenced by a live build elsewhere, and TASK_REFERENCED leaves
  `latest_status_build_id` alone. `POST /builds/{id}/skip-blocked` is the
  operation for pending work.

Default off because it is a behaviour change for existing callers, and the
SDK's fail-fast path already cancels its own running tasks.

### 2 + 3. `POST /api/v1/builds/bulk-cancel`

Bulk cancel and "reap builds idle beyond a threshold" are the same
operation with different filters, so they are **one endpoint** rather than
two near-identical code paths. A follow-up `stardag builds cleanup` sits
on top of it.

Request body:

```jsonc
{
  "build_ids": ["…"],          // UUID[]|null  — explicit selection
  "idle_for_seconds": 86400,   // int|null, >= 60 — the reaper filter
  "reactive_app_name": "…",    // string|null — implies include_reactive
  "include_reactive": false,   // bool
  "cascade": true,             // bool — default TRUE here (see below)
  "dry_run": false,            // bool
  "limit": 100,                // int 1..500 — caps the WRITE set only
  "reason": "…"                // string|null — recorded on the event
}
```

At least one of `build_ids` / `idle_for_seconds` is required (422
otherwise): an unqualified "cancel everything running here" is not a
cleanup operation. `idle_for_seconds` has a floor of 60s — a threshold
small enough to race a live build is a foot-gun.

Response:

```jsonc
{
  "dry_run": false,
  "builds": [
    {
      "build_id": "…",                            // string (UUID)
      "name": "brave-tiger-42",                   // string
      "last_activity_at": "2026-01-01T00:00:00Z", // string|null — the signal selected on
      "reactive_app_name": null,                  // string|null
      "cascaded_task_ids": ["…"]                  // string[]
    }
  ],
  "build_count": 1,        // int
  "task_count": 1,         // int — claims released
  "skipped": {             // {build_id: reason} — only for explicit build_ids
    "…": "not_running"     // "not_found" | "not_running" | "reactive" | "not_idle" | "limit_reached"
  },
  "truncated": false       // bool — more matched than `limit`; call again
}
```

`cascade` defaults to **true** here (unlike the single-build cancel, where
it defaults off for compatibility): releasing leaked claims and slots is
the entire point of a cleanup pass, and a build cancelled without it leaves
behind exactly the problem the endpoint exists to fix.

`not_found` covers both unknown ids and ids belonging to another
environment — indistinguishable on purpose, so the endpoint cannot be used
to probe for build ids elsewhere.

Only builds whose derived status is RUNNING are ever touched, whatever the
filter says. That is what makes the call idempotent: safe to retry, safe to
run on a timer, safe for two replicas concurrently.

Auth: destructive, so the JWT/UI path requires the workspace **ADMIN** role;
API keys (environment-scoped machine credentials) stay unrestricted — the
same gate as concurrency-limit eviction.

### 4. `idle_for_seconds` on `GET /api/v1/builds`

Added after a downstream finding: a client that wants "builds idle for a
day" had no server-side filter, so it fetched one wide window, filtered in
the browser, and warned that older matches might be missing — the same
silent-truncation failure this PR refuses for the reaper, and for the same
reason. The dropped builds are the oldest ones, which are exactly what the
query exists to find.

| param | type | default | notes |
| --- | --- | --- | --- |
| `page` | int ≥ 1 | 1 | unchanged |
| `page_size` | int 1–100 | 20 | unchanged |
| `reactive_app_name` | string | – | unchanged |
| `status` | `BuildStatus` | – | unchanged on its own; see the interaction below |
| **`idle_for_seconds`** | int **≥ 60** | – | no activity for at least this long |

Response shape is unchanged (`BuildListResponse`); every build carries the
`last_active_at` / `last_activity_at` fields described above.

- **Same definition of idle as the reaper, from one shared function.** The
  three-conjunct AND moved out of `select_cancellable_builds` into
  `idle_filters()` in `services/build_cleanup.py`, used by both call sites.
  This endpoint is how an operator previews what `bulk-cancel` will do, and
  a preview that disagrees with the action is worse than no preview. Same
  `ge=60` floor for the same reason.
- **`total` is exact.** A real SQL predicate means `COUNT(*)` over it is
  exact and pagination is server-side and unbounded — deliberately *not*
  extending the derived-`status` path's "matches within the scanned window"
  wart to this filter.
- **Ordering flips to stalest-first** when the filter is given (default
  stays most-recently-active-first, so a resumed build still jumps to the
  top). The sort key is the shared `STALEST_FIRST_ORDER`, i.e. the same
  documented `last_active_at` *proxy* the reaper uses — sorting on the full
  `last_activity_at` signal would forfeit
  `ix_builds_environment_last_active` and sort the whole candidate set for a
  max over a correlated aggregate. Consequence worth knowing for the UI:
  the rendered `last_activity_at` column is not guaranteed perfectly
  monotonic. Correctness lives in the filter; ordering only decides which of
  the already-correctly-filtered builds come first.

**Interaction with `status`:**

- `status=running` + `idle_for_seconds` → accepted, and `status` then
  resolves via the SQL `build_is_running()` predicate instead of the Python
  scan over `_STATUS_FILTER_SCAN_CAP = 500` most-recently-active
  candidates. That combination is *the* abandoned-build query; answering it
  over a most-recently-active window would drop precisely the builds it
  targets. Exact `total`, unbounded pagination.
- any other `status` + `idle_for_seconds` → **422**, with a message
  explaining that only `running` has a SQL predicate and that the others
  would make `total` and pagination approximate without saying so.
  Returning a plausible-looking wrong number is worse than refusing.
- `status` on its own → completely unchanged, including its existing
  window-scan behaviour and its approximate `total`. Generalising the
  derived-status→SQL translation to the terminal statuses is possible
  (each is "the most recent build-level event is of type X") but is a
  separate change.

### Which idleness signal, and why

**`Build.last_active_at` is not an activity signal, and using it would
cancel live work.** Its model docstring says only build-level lifecycle
events touch it, and every `_touch_build_last_active` call site confirms
that: `/resume`, `/complete`, `/fail`, `/cancel`, `/exit-early`, and
`/roots`. Task events deliberately skip it so worker traffic doesn't
serialise on the build row. A build that has been busily running tasks for
three days therefore still shows its BUILD_STARTED timestamp there.

The reaper measures **`last_activity_at`** instead: the newest of

1. `max(events.created_at)` over the build's **entire** event stream — task
   registrations, starts, completions, failures included. Served by
   `ix_events_build_created` as a one-row backward index scan.
2. `Build.last_active_at` — lifecycle transitions, plus appending roots,
   which writes no event at all.
3. `Build.needs_tick_at` — a pending scheduler wake-up. Set by
   `POST /builds/{id}/notify` (a worker reporting it finished something)
   and cleared by the tick that consumes it. While it is set, somebody is
   waiting on a scheduler that hasn't run — the opposite of abandoned.

In SQL the filter is an AND of "each input is old", which is equivalent to
`max(inputs) < cutoff` without needing `GREATEST()` (absent on SQLite).

`BuildResponse` now exposes **both** `last_active_at` (the ordering column,
already on the model but never in the response) and `last_activity_at` (the
derived signal), so the UI can render a staleness column showing exactly
the number the reaper will act on. Both land on every build response —
create, get, list, and each lifecycle endpoint.

All three consumers — the reaper, `GET /builds?idle_for_seconds=`, and the
`last_activity_at` each response reports — read the same definition from
`idle_filters()` / `last_activity_at()`.

**Reactive builds are excluded by default.** A reactive build is quiet
between ticks by design and its ticks emit no events when there is nothing
to do, so silence is not evidence of abandonment — and it already has a
watchdog for the wedged case. `include_reactive` / `reactive_app_name` opt
in, for when you know the owning app is gone.

**Ties are read as not-running.** A terminal build-level event sharing a
timestamp with a start/resume is treated as terminal, so the reaper leaves
the build alone. `get_build_status` would resolve such a tie arbitrarily;
where the two can disagree, the conservative reading wins.

### Cost

The RUNNING test is a SQL predicate — two correlated aggregates over
`ix_events_build_created`, comparing the newest start/resume against the
newest terminal — **not** the bounded Python scan behind
`GET /builds?status=` (`_STATUS_FILTER_SCAN_CAP = 500`). A reaper
inheriting that cap would systematically miss the oldest stale builds,
which are exactly the ones it exists to find, and it fails silently as an
environment ages (see #208 A3).

The idle filter puts an index-backed `Build.last_active_at < cutoff`
conjunct in front of the aggregates. It is logically redundant but
load-bearing for the plan, and sound: the real signal is a max() that
*includes* that column, so the index-backed conjunct can never exclude a
build the full predicate would keep. `ix_builds_environment_last_active`
therefore narrows the candidate set before any aggregate runs.

Writes are capped by `limit` (with `truncated` telling you to call again);
the scan is not. The whole batch commits in one transaction, so a crash
cannot leave a cancelled build still holding its claims.

Entity-limit accounting pre-checks the BUILD_CANCELLED events only — the
cascade count isn't known until the task rows are locked, and re-selecting
them up front would double the query cost of every sweep. Cascaded events
are recorded against the quota afterwards, so a workspace near its ceiling
is stopped on the *next* call. Acceptable for an admin path whose write set
is already capped.

### Optional unattended sweep

`STARDAG_API_REAPER_*` settings, **off by default**, run the same operation
on a timer inside the API process:

| env var | default |
| --- | --- |
| `STARDAG_API_REAPER_ENABLED` | `false` |
| `STARDAG_API_REAPER_INTERVAL_SECONDS` | `900` |
| `STARDAG_API_REAPER_IDLE_FOR_SECONDS` | `86400` |
| `STARDAG_API_REAPER_INCLUDE_REACTIVE` | `false` |
| `STARDAG_API_REAPER_CASCADE` | `true` |
| `STARDAG_API_REAPER_MAX_BUILDS_PER_SWEEP` | `100` |

**Multi-replica caveat:** every replica runs its own timer; there is no
leader election. Cancelling an already-terminal build is a no-op, so
concurrent sweeps are wasteful (duplicated scan, harmless race on the same
rows), not wrong. With more than a couple of replicas, prefer an external
scheduler calling the endpoint once. It waits a full interval before the
first sweep so a crash-looping process never reaps, and swallows per-sweep
exceptions so a transient DB error can't kill the loop.

---

## Refactor riding along

The nine copies of the ~20-line BuildResponse assembly block in
`routes/builds.py` now go through the existing `_build_to_response`
helper. Not cosmetic: it is what makes the two liveness fields land
consistently on every build response instead of on the two endpoints that
already used the helper. `list_builds` fetches the page's activity
timestamps in one grouped query rather than one per build.

## Tests

`app/stardag-api/tests/test_build_cleanup.py` (new) and additions to
`test_tasks.py`. The single most important one is
`test_build_with_recent_task_activity_is_not_reaped`: a build whose
`last_active_at` and build-level events are two days old but which
registered and started a task seconds ago must survive, while a genuinely
abandoned build in the same sweep is cancelled.

Also covered: multi-value status filtering; the `status_older_than` strict
boundary; oldest-claim-first ordering; the new `TaskResponse` fields;
cascade releasing a claim *and* freeing the concurrency-limit slot (proved
by a second build acquiring it under a cap of 1); cascade refusing to touch
another build's running task and refusing to touch PENDING; bulk cancel by
ids; idempotence and the `skipped` reasons; dry run changing nothing while
reporting exactly what the real run then does; `needs_tick_at` blocking a
reap; resumed builds counting as RUNNING while completed/failed ones are
skipped; reactive exclusion and opt-in; the `limit`/`truncated` drain;
environment isolation on every new endpoint; the ADMIN gate and the
unauthenticated path. For the `GET /builds` idle filter
specifically: exactness of `total` and a server-side page partition over
the matches; stalest-first ordering (and the default order left alone); the
same recent-task-activity guard as the reaper, asserted to agree with a
`bulk-cancel` dry run on the same data; `needs_tick_at` blocking a match;
the shared `ge=60` floor rejecting on both endpoints; the 422 on a
non-`running` status; and `status=running` + `idle_for_seconds` returning
the old builds with the scan cap monkeypatched down to 2, which the bare
`status` filter still truncates. Postgres variants (`pg_client`) exercise the
correlated-aggregate predicate and the indexed task query on the dialect
they are written for.

Fixture note: `as_environment_b` and the role switcher moved from
`test_builds.py` into `conftest.py` (first commit) since both boundaries
now need testing in more than one module.

## Migration head

`11716b1f4c1a` revises `792fb8ab3c3c`. Note that #211
(`feat/persist-reactive-tick-summaries`) also adds a migration revising
`792fb8ab3c3c`, so **expect an alembic multi-head situation when both
merge** — it needs a merge revision at that point, not a change here.

## Verification

```
cd app/stardag-api && uv run --no-sync pytest tests/ -q
# 406 passed, 22 skipped   (SQLite)
STARDAG_API_TEST_DATABASE_URL=… uv run --no-sync pytest tests/ -q
# 427 passed, 1 skipped    (Postgres 16)

uvx pre-commit run --files <every changed file>
# prettier / ruff / ruff-format / pyright-stardag-api — all pass
```


andhus:
Copilot review triaged — four findings, all accepted.

1. **Alembic header `Revises:` disagrees with `down_revision`.** Real; the header is stale after the chain was relinearised. Confusing exactly when someone is debugging migration order.
2. **`limit_reached` mislabelling.** The sharpest of the four: when `truncated`, any visible RUNNING non-reactive build that was not selected is reported as `limit_reached` even if it was skipped for *not being idle*. That is a wrong answer to "why was this id not acted on".
3. **`skipped` docs omit `limit_reached`.** Real.
4. **Entity-limit accounting.** Real — the build-level `BUILD_CANCELLED` event is not counted alongside the cascade, and since `check_entity_creation_limit` does not reserve, `1 + len(cascaded)` can exceed the limit.

andhus:
**Now resolved** — all four.

- `limit_reached` no longer swallows `not_idle`: eligibility is determined under the request's own filters and checked before truncation, so the one skip reason that means "call again" is never handed to a build no retry will select. Regression test included (it fails without the fix).
- Cascade entity-limit now counts the BUILD_CANCELLED event (`len(cascade) + 1`).
- `skipped` documents `limit_reached`.
- Migration header matches `down_revision`; audited the other migrations, no further drift.

PR #214, opened 2026-08-07
fix(reactive): wait for cross-build blockers instead of failing the build (#208 A1/A2)

andhus:
SDK side of **A1** and **A2** from #208. `lib/` and `docs/` only. The
server half (#210) is what makes the frontier carry the information this
consumes; against an older server every behaviour here degrades to exactly
what ships today.

## A1 — a blocker in another build no longer kills the build

A tick read `not frontier.actionable and running == 0` as "this build
cannot progress" and failed it. Both signals are **build-scoped** while
dependency gating is **environment-global**: task rows and their dependency
edges are per environment, so an upstream some other build left RUNNING
gates this build's downstream tasks while contributing to neither this
build's `running` nor its `status_counts`. It need not be in this build's
task set at all — a dynamic dependency registered under an earlier build is
the usual case. Result: healthy builds died within seconds of triggering,
with an error reporting only status counts, which points nowhere near the
cause.

Terminal detection now classifies the frontier's `blocked_by_external` on
two questions — is the blocker in this build's own task set, and **is
anyone going to move it**:

| `blocking_in_build` | `blocking_status` | owning build (`blocking_status_build_id`) | action |
| --- | --- | --- | --- |
| `true` | any | not consulted | **ignored** for the wait/fail decision — the blocker is in this build's own task set and is already modelled by `actionable` / `running` / `status_counts`. Still named in the message, since the counts alone never said which task was holding things up. |
| `false` | `running` | not consulted | **wait** — another build is executing it; a completion frees this build. Treated exactly like the existing `denied_this_round > 0` case: return `None`, don't fail. |
| `false` | not `running` | live (`running`, or `pending` — it may yet start) | **wait** — that build is going to schedule it. |
| `false` | not `running` | terminal (`completed` / `failed` / `cancelled`) | **fail now** — nothing is going to move it and this build never will. |
| `false` | not `running` | `None` (no build owns its status) | **fail now**, without a lookup — no status-moving event was ever recorded against the blocker, so there is no evidence anyone intends to run it, and no build to ask. |
| `false` | not `running` | unresolvable (deleted build, unreachable registry, or a server that doesn't report `status`) | **fail now** — unknown is not evidence of life, and a silent indefinite hang is the failure mode #208 exists to kill. |

Both waits are bounded (below). `fatal` outvotes `waiting`: one blocker
nothing will ever run means the build cannot complete, whatever else it is
also waiting on.

The owner-liveness row is the fix for the gap an earlier revision of this PR
left open — an out-of-build blocker that was merely *pending* under a
perfectly healthy concurrent build would have been failed, which is a *new*
spurious failure, the exact class of bug being fixed. The frontier already
names the owning build and `GET /builds/{id}` already returns its derived
status; the SDK was discarding it because `BuildInfo` didn't model the field
and `StardagBaseModel` ignores extras. `BuildInfo.status: str | None = None`
now models it (defaulted, so older servers and custom registries
deserialize unchanged). `build_get_aio` already existed on both
`RegistryABC` and `APIRegistry`, so no new registry surface was needed.

**Cost:** the owner status is resolved **once per distinct owning build per
classification pass** — a wide DAG stalled behind one build yields one
blocker entry per blocked edge, every one naming the same owner, so without
the memo this would be N requests for one answer. Deliberately not cached
across passes, so a later pass can see the owner go terminal. And the whole
lookup only happens on the stalled path, since the frontier populates
`blocked_by_external` solely when the build has nothing actionable and
nothing running — a healthy build issues **zero** extra requests however
often it polls. Stated in a comment so nobody "optimises" it away later.
Lookup failures are swallowed with a warning rather than escaping the tick:
this path decides a build's fate and must produce a precise failure, not a
traceback.

### The bounded wait

New `TickConfig.stale_external_blocker_seconds`, **default 21600.0 (6 h)**,
measured on the blocker's own `blocking_status_at` — how long it has sat in
that status — not on tick-local time. A tick is short-lived and the watchdog
re-ticks forever, so a tick-local bound would never expire; the build would
just hang quietly, which is the failure mode #208 is about. It applies to
both wait cases: a live owner is not a licence to wait forever, since a
build that has left a task pending for a day is not making progress on it.

Deliberately generous. Waiting is the cheap direction: the blocker's own
build has far better information about it (a live executor ref it can probe)
and its completion wakes this build within a watchdog period, whereas
failing a build whose blocker was merely slow is a regression of the exact
spurious-failure bug being fixed. The bound is a backstop for a blocker
nobody will ever move, not a scheduling deadline. `None` waits indefinitely
(with a per-tick log line saying so).

**`blocking_status_at is None`** (a task row predating server-side status
denormalisation): the wait cannot be aged, so the blocker is waited on
regardless of the bound, and the log line says the wait is unbounded.
Failing on missing information would reintroduce the very failures this
removes. Documented on the config field and covered by a test.

Not added to `_TICK_KWARGS_ALLOWED` (the per-build tick kwargs persisted in
the registry), matching `stale_running_no_ref_seconds`, which is likewise
app-level only.

### The two messages

They read differently on purpose: waiting because another *live* build owns
the blocker is a different operational situation from failing because none
does, and the operator needs to be told which. Each blocker also carries the
verdict's reason (`_BlockerVerdict`), so the description says *why* it
landed where it did.

**Waiting** (`INFO`), verbatim, from a test — note the executing/queued
split and the unbounded-wait note:

```
Build 66a2… has nothing runnable or running of its own but is waiting on 2
upstream task(s) owned by other builds (2 queued in a build that is still
live); waiting rather than failing (a blocker carries no status timestamp,
so its wait cannot be bounded): task 132e8369-… is blocked by
pipelines.Ingest (blocking-task-id), PENDING for 60s under build 66a29a9c-…
— its owning build is still running; task 9896df36-… is blocked by
pipelines.Transform (second-blocker), PENDING under build 66a29a9c-… — its
owning build is still running
```

**Failing** (`ERROR`, and the build's `error_message`), verbatim:

```
Build cannot progress: it has nothing runnable or running of its own, and 1
of its task(s) are blocked by an upstream owned by another build that no
live build is going to run (status counts: {'pending': 1}). Blocked by: task
132e8369-… is blocked by pipelines.Ingest (blocking-task-id), SUSPENDED for
60s under build e85c3f0e-… — its owning build is completed. Retry the
blocking task under the build that owns it (POST
/api/v1/builds/{build_id}/tasks/{task_id}/retry) to reset it to pending —
this now works from suspended as well as from failed/cancelled/skipped —
then re-trigger this build.
```

The cancel-first clause ("A blocker stuck RUNNING holds an execution claim
no retry can take…") is appended only when a fatal blocker is RUNNING. The
message names at most 5 blockers and appends "and N more"; when the server
set `blocked_by_external_truncated` it says so rather than implying the list
is exhaustive. The generic stuck-build failure keeps its old wording and now
appends any **in-build** blockers, which the status counts never identified.

`TickSummary` gains three flat counters (a later PR posts the summary to
the registry): **`external_blockers`**, **`external_blockers_waited`**,
**`external_blockers_fatal`**. The first counts every entry reported while
the build looked stalled, including in-build ones; the two splits cover
only out-of-build entries, so they do not add up — documented on the
fields. Existing field names untouched.

## A2 — `suspended` is retryable

`_RETRYABLE_STATUSES` gains `"suspended"`, so a re-trigger's
`discover_and_register_aio(..., retry_failed=True)` resets a task left
SUSPENDED — its execution registered dynamic dependencies, yielded and
returned, and then the orchestrator died or the build was cancelled —
instead of skipping it. Previously such a task was permanently
unschedulable, and the only escape was cancelling it purely to reach a
status that *was* retryable.

Safe precisely because a SUSPENDED task has **no live execution**: the
suspension means the execution yielded and *returned*, so resetting it
cannot orphan a running worker. Stated beside the constant, where someone
wondering whether this can strand an execution will look.

The other caller of `discover_and_register_aio` is a worker registering its
dynamically yielded deps; it takes the default `retry_failed=False` and so
never retries anything — verified, noted beside the constant, and pinned by
a test, since that is what stops a suspending worker from resetting its own
task out from under itself. `running` remains deliberately non-retryable:
it holds a live execution claim, and releasing that is cancellation.

## Behaviour change worth knowing: `stale_running_no_ref_seconds` now works

#210 also fixed `FrontierTaskRef.latest_status_at`, which was declared and
documented as the input to scheduler staleness bounds but never populated —
it silently serialised as `null` on every ref, leaving the guard dead.
Against a matching server, a task stuck RUNNING without an executor ref for
longer than the bound (**default 30 minutes**) is now failed as intended.
Raise the bound if you run long ref-less tasks — e.g. a resident build
sharing tasks with a concurrently ticking reactive one. No SDK change was
needed for this; it starts working on upgrade, which is why it is called
out here.

## Model

`FrontierExternalBlocker` in `stardag.registry` (exported alongside
`FrontierTaskRef`), plus `BuildFrontier.blocked_by_external` /
`.blocked_by_external_truncated` and `BuildInfo.status`, all defaulted so an
older server's response deserializes unchanged. The frontier field comment
records the semantics that matter most to a consumer: the server computes
the list **only** when the build has nothing actionable and nothing running,
so an **empty list means "not externally blocked, OR not stalled"** and must
never be read as proof that no external blocker exists.

## Tests

`FakeReactiveRegistry` gained the API's two scopes — an environment-global
status map plus `not_in_build`, the ids that exist in the environment but
not in this build's task set — computes `blocked_by_external` under the same
"only when stalled" gate as the server, and now serves per-build derived
status from `build_get_aio` (recording every call, so memoisation is
assertable). New `TestExternalBlockers` covers:

- the A1 repro: a build whose task waits on an upstream RUNNING in another
  build now lingers out with the build still RUNNING, where it used to be
  failed;
- a non-RUNNING blocker under a `running` **and** under a `pending` owner →
  waits; the same blocker past the bound → fails, message naming the bound;
- a terminal owner → fails; `blocking_status_build_id is None` → fails with
  no lookup at all; a 404ing lookup → fails without propagating; a server
  not reporting `status` → fails the same way and says so;
- **memoisation**: two blocker entries sharing one owner produce exactly one
  `build_get` call;
- the RUNNING-blocker bound expiring (a blocker RUNNING for a day) and the
  control case (the same age under a larger bound still waits);
- an in-build blocker that buys no wait but does get named;
- an unbounded wait: no `blocking_status_at` even with the bound set to 0,
  and `stale_external_blocker_seconds=None`;
- a fatal blocker outvoting a waitable one, and only the fatal one named;
- the truncation flag reaching the message;
- an older server (`serves_blocked_by_external = False`) producing exactly
  the pre-fix failure and zero counters;
- a healthy build recording no blockers at all;
- the end of the wait: the owning build completes the blocker, and the same
  lingering tick schedules the task it would previously have failed over.

Plus, in `TestRetryPath`: a suspended task reset by `retry_failed=True` and
the build running to completion, and a worker-style call (`retry_failed`
defaulted) leaving `suspended` untouched with no retry call at all.

`FakeReactiveRegistry.task_retry_aio` now applies `_RETRYABLE_STATUSES`
directly instead of its own copy of the tuple, so the fake cannot drift
from the code under test.

## Verification

From `lib/stardag`:

```
uv run pytest tests -q -m "not modal_live"   # 1053 passed, 29 deselected
uv run pyright src tests                     # 0 errors, 0 warnings
```

plus `pre-commit` (prettier, ruff, ruff-format, pyright) on every changed
file. The `modal_live` tier was not run.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -37,12 +37,14 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
   about classes the patterns don't cover, naming the pattern to add;
   `require_pickle_free=True` turns that fallback into a hard error.
 
-  Resident (non-reactive) builds are unaffected, and `task_modules=[]`
-  behaves exactly as before. **Upgrade note for reactive users:** because
-  the default _infers_ patterns, a reactive trigger from the upgraded SDK
-  may skip pickles that the currently _deployed_ tick cannot compensate
-  for — redeploy the app before triggering (reactive mode already requires
-  the deployed app and the SDK to match), or pass `task_modules=[]`.
+  Skipping pickles requires declaring `task_modules` explicitly; the
+  inferred default only drives the coverage warning. Upgrading stardag
+  therefore changes nothing on its own — a newer SDK triggering against an
+  app deployed by an older one still writes pickles, because eliding them
+  would depend on a baked-in module list that deployment does not have.
+  Resident (non-reactive) builds are unaffected either way, and
+  `task_modules=[]` behaves exactly as before. **Redeploy the app whenever
+  you change `task_modules`**, before triggering.
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,44 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
 ## [Unreleased]
 
+### SDK
+
+- **`StardagApp(task_modules=[...])`: declare the modules whose import
+  registers your task classes, so reactive scheduler ticks can rebuild
+  tasks from registry data instead of pickles.** A tick reconstructs task
+  objects from the registry's stored payload, which resolves a class
+  through the polymorphic registry — populated only as a side effect of
+  importing the defining module. Without a declaration, whatever a tick
+  container happens to import is arbitrary, so the build task store's
+  pickles were load-bearing (and needed target-root write access at
+  trigger time, and were invalidated by every redeploy).
+
+  Patterns are exact modules (`"my_pkg.tasks.ingest"`) or trailing
+  recursive wildcards (`"my_pkg.tasks.*"`); the default infers the root
+  package of the module defining the app, and `[]` opts out. They are
+  expanded to a concrete module list at deploy time (without importing
+  submodules) and baked into the deployed tick, so **adding or moving task
+  classes requires a redeploy**; `stardag modal deploy` reports the
+  expansion (`--no-check-task-modules` skips the warn-only local import
+  check). Task modules are imported in every tick container, so keep heavy
+  runtime dependencies inside `run()` rather than at module scope.
+
+  With the declaration in place, a reactive trigger writes **no pickle**
+  for any task whose class is covered and whose payload round-trips to the
+  same task id — a fully covered build needs no target-root write access
+  at all. Everything else keeps its pickle exactly as before, including
+  `AliasTask` payloads (pickled `loads_type`, never auto-unpickled from
+  registry data by design) and non-importable classes. The trigger warns
+  about classes the patterns don't cover, naming the pattern to add;
+  `require_pickle_free=True` turns that fallback into a hard error.
+
+  Resident (non-reactive) builds are unaffected, and `task_modules=[]`
+  behaves exactly as before. **Upgrade note for reactive users:** because
+  the default _infers_ patterns, a reactive trigger from the upgraded SDK
+  may skip pickles that the currently _deployed_ tick cannot compensate
+  for — redeploy the app before triggering (reactive mode already requires
+  the deployed app and the SDK to match), or pass `task_modules=[]`.
+
 ### Fixed
 
 - **The reactive watchdog now asks the registry only for the RUNNING builds
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,23 +57,33 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
   a dynamic dependency registered under an earlier build, so it is not in
   the new build's task set at all.
 
-  A tick now reads the frontier's blocking upstreams and acts on them: a
-  blocker **running under another build** is waited out (like a busy
-  concurrency-limit slot — its completion wakes this scheduler), while a
-  blocker **nobody is executing** fails the build immediately, naming the
-  task, its namespace/name, its status, how long it has been in it, the
-  build that owns it, and how to release it. A blocker inside this build's
-  own task set is unchanged: it is already visible in the counts, and still
-  fails the build when it can never run.
+  A tick now reads the frontier's blocking upstreams and asks, for each,
+  whether anyone is going to move it. A blocker **another build is
+  executing** is waited out (like a busy concurrency-limit slot — its
+  completion wakes this scheduler), and so is one **a still-live build has
+  yet to schedule**: that build is going to run it, and failing here would
+  only trade one spurious failure for another. A blocker **no live build is
+  going to run** — its owning build has gone terminal, no build owns its
+  status, or that status could not be resolved — fails the build
+  immediately, naming the task, its namespace/name, its status, how long it
+  has been in it, the build that owns it, why that owner will not move it,
+  and how to release it. A blocker inside this build's own task set is
+  unchanged: it is already visible in the counts, and still fails the build
+  when it can never run.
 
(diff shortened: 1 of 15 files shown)


andhus:
Copilot review addressed — the finding was right and the fix changed a test that was pinning the wrong behaviour.

Both remedies quoted in that message are addressed to a *build*, so for a blocker whose owning build was never recorded there is no id to substitute into either URL. The guidance was the shape of a remedy rather than one, in exactly the case it exists for.

Owned and unowned blockers are now described separately. The unowned case says what can actually be done: look the task up (registry UI or `stardag tasks list`) to find the build holding its claim, or re-run it under a new build if the status predates claim recording.

`test_blocker_with_no_status_owning_build_fails` asserted `/retry` appeared in that message — i.e. it was pinning the non-actionable output. Updated to assert the new guidance instead.

PR #215, opened 2026-08-07
feat(ui): Builds view with staleness, batch selection and bulk cleanup (#208 B3/B5)

andhus:
Frontend half of the stale-build cleanup work in #208 — the operator side
of **B3** (nothing terminates abandoned builds, and there is no bulk
operation) and the build-list part of **B5** (the UI cannot explain a
stalled build). Builds on the API branch it targets.

The motivating shape of the problem: an environment accumulated dozens of
builds left `RUNNING` forever by orchestrators that died without emitting
a terminal event, each still holding the execution claims and
concurrency-limit slots its tasks had when it vanished. Nothing in the UI
showed which builds those were, and cleaning up meant hand-rolled API
calls.

## What changed

**"Home" is now "Builds".** The nav item said "Home" but the view has
always been a builds list — its own breadcrumb reads "Builds", and
`NavItem` already carried an unused `"builds"` member. Label, nav id and
`HomePage` renamed; `/` is unchanged as the route.

**The builds list is a real table.** Columns: selection, status, build
(with the short-commit chip as before, plus a reactive-app chip that
doubles as a filter shortcut), description, duration, last activity,
created. Three filters — `status`, `reactive_app_name` and
`idle_for_seconds` — all server-side, all visible in the toolbar, all
clearable. The empty state distinguishes "no builds yet" from "no builds
match these filters".

**`last_activity_at`, not `last_active_at`.** These are two different
numbers and the difference matters here. `last_active_at` is what orders
the list by default, but it is bumped by build-level *lifecycle*
transitions only — task events deliberately skip it — so a build that has
been running tasks for three days still reports its `BUILD_STARTED`
timestamp. Surfacing it would tell an operator that a busy build is idle.
The table shows and filters on `last_activity_at`: the newest of the
build's whole event stream, its `last_active_at`, and any pending
scheduler wake-up — exactly what the reaper measures. Rendered as a
relative time with the absolute time in a `title`, and running builds
quiet for over a day are flagged in place without any filter being set.

**Idleness is filtered by the server, with the reaper's own predicate.**
`GET /builds?idle_for_seconds=` and `POST /builds/bulk-cancel` share one
SQL predicate on the API side, so the set this table shows and the set
"Clean up idle builds" acts on are the same by construction rather than
because two implementations happen to agree. `total` is an exact `COUNT`
and pagination is unbounded, so there is no scan-window caveat to make.
Ordering also comes from the server — stalest-first with the filter,
most-recently-active-first without — and the page is rendered as
returned, never re-sorted client-side.

One honesty note carried into the UI rather than hidden: the server sorts
stalest-first on `last_active_at`, an index-backed proxy for the composed
`last_activity_at` the column renders. They nearly always agree, so the
column runs *close to* — not strictly — oldest-first, and the "Last
activity" tooltip says exactly that.

`idle_for_seconds` alongside any status but `running` is a 422 (only
RUNNING has a SQL predicate). That combination is unreachable from both
directions rather than left to fail: incompatible status options are
disabled, and labelled "n/a with an idle filter", while an idle filter is
set; the idle select is disabled while an incompatible status is
selected. Both carry a tooltip explaining why, and neither control
silently rewrites the other.

**Bulk cancel shows its consequences before committing.** Selecting rows
reveals a bulk bar; confirming opens a dialog that immediately issues the
endpoint's `dry_run` and reports what a real call would do — how many
builds, how many task claims a cascade would release, and every build the
server will skip *with the reason spelled out in prose* (`not_running`,
`reactive`, `not_idle`, `limit_reached`, `not_found`), plus `truncated`.
The dry run re-runs whenever cascade or include-reactive changes, so
"3 builds skipped: reactive" is actionable rather than a dead end. Only
then does the confirm button apply the identical filter with
`dry_run: false`, and the outcome lands in a banner above the refreshed
table.

"Clean up idle builds…" runs the same dialog in `idle_for_seconds` mode,
enabled only once the threshold is explicit in the filters — so the
destructive action is always tied to a visible number, and to the exact
set the table is already showing.

**Single-build cascade.** `POST /builds/{id}/cancel` now takes `cascade`,
so the build view's override menu gains a second entry, "Cancel & Release
Claims", next to the untouched plain Cancel. Separate action rather than a
changed default: existing behaviour is preserved and the difference — one
writes a build event, the other actually releases the claims — is visible
in the menu. It reports how many claims were released.

**Selection across pagination:** scoped to the visible page, cleared on
page, filter or environment change. Carrying off-screen rows into a bulk
cancel is exactly the kind of hidden state that makes a destructive action
untrustworthy, so the hook stores the selection together with the key it
was made under and a selection made under a different key reads as empty.

**Admin gating:** bulk cancel requires the workspace admin role on the JWT
path (same gate as concurrency-limit eviction), so non-admins get no
selection column and no dead buttons, with a note in the toolbar saying
why. Filters and navigation are unchanged for everyone.

## Shared primitives

The codebase had no checkbox anywhere, no shared selection, and no
confirmation beyond `window.confirm` plus an inline red div. Introduced
once, small, no new dependency, and deliberately generic so the follow-up
work on this issue can reuse them:

| Primitive | Purpose |
| --- | --- |
| `ui/Checkbox` | The one checkbox, with the indeterminate state a "select all on page" header needs and a required accessible name. |
| `hooks/useRowSelection` | Selection for a paginated table; derived reset, no effect. |
| `ui/BulkActionBar` | Appears only when a selection exists; announces the count via `aria-live`. |
| `ui/ConfirmDialog` | Composed from the existing `Modal` (which gains an optional `maxWidthClass`), free-form body so a bulk action can render its preview inside the dialog that commits it. Focus lands on Cancel. |
| `ui/ResultBanner` | Inline outcome/error, `role=status` or `role=alert` by tone. |
| `utils/time` | The relative/absolute/duration formatting that was private to `BuildsList`, plus `secondsSince`. |

Existing `window.confirm` call sites are left alone; the dialog is used
only for the new destructive actions.

## Notes

Every new surface carries explicit `dark:` pairs. Rows are `<tr>`s with a
whole-row click for the mouse, but the build name is a real `<button>`, so
the row stays keyboard-reachable and accessibly named now that it is no
longer a full-width `<button>`; a click anywhere in the checkbox cell
selects and never navigates. Data fetching follows the existing
conventions — `useCallback` loader plus `useEffect`, refetch after mutate,
the epoch-ref stale-response guard on both the list and the dry run, and
the pre-existing `lastFetchedEnvIdRef` environment-change guard is
preserved. Types and API calls added by hand as usual, including a helper
that surfaces FastAPI's `detail` rather than "Unprocessable Entity".

`Modal` picks up a `TODO(a11y)` naming its missing focus trap — Escape
closes and the body is scroll-locked, but Tab walks out of the dialog and
focus is not restored on close. Shared behaviour inherited by every
caller, so it wants its own change rather than riding along here.

## Tests

22 new cases (`BuildsList.test.tsx`, `ui/Checkbox.test.tsx`): the
activity-vs-lifecycle distinction (asserting the lifecycle timestamp is
*not* rendered), filter round-trips including `idle_for_seconds`, proof
that the page is neither re-sorted nor re-filtered locally, the
status-plus-idleness pair being unreachable in both directions, selection
/ select-all / indeterminate header, the checkbox-does-not-navigate rule,
keyboard activation of a row, selection clearing across pagination, the
dry-run-then-apply flow including "nothing written yet", skip reasons and
truncation, the cascade toggle re-running the preview, admin gating, and
the empty/error states. `Sidebar.test.tsx` updated for the rename.

```
npx vitest run   → 12 files, 118 tests passed (96 before)
npx tsc -b       → clean
npx eslint src/  → clean
uvx pre-commit run --files <all changed>  → prettier / eslint / tsc pass
npx vite build   → clean
```

## Not in scope

The "why is this build not progressing?" panel, blocker visibility on the
build detail view, and the task-explorer claim work are a follow-up; the
primitives above are shaped for it.

### DIFF
--- a/app/stardag-ui/src/App.tsx
+++ b/app/stardag-ui/src/App.tsx
@@ -111,21 +111,21 @@ interface SidebarStateProps {
   onToggleSidebar: () => void;
 }
 
-// Home page with builds list
-interface HomePageProps extends SidebarStateProps {
+// Builds list page (the app root view)
+interface BuildsPageProps extends SidebarStateProps {
   onNavigate: (item: NavItem) => void;
   onSelectBuild: (buildId: string) => void;
 }
 
-function HomePage({
+function BuildsPage({
   onNavigate,
   onSelectBuild,
   sidebarCollapsed,
   onToggleSidebar,
-}: HomePageProps) {
+}: BuildsPageProps) {
   return (
     <MainLayout
-      activeNav="home"
+      activeNav="builds"
       onNavigate={onNavigate}
       sidebarCollapsed={sidebarCollapsed}
       onToggleSidebar={onToggleSidebar}
@@ -153,7 +153,7 @@ function BuildPage({
 }: BuildPageProps) {
   return (
     <MainLayout
-      activeNav="home"
+      activeNav="builds"
       onNavigate={onNavigate}
       sidebarCollapsed={sidebarCollapsed}
       onToggleSidebar={onToggleSidebar}
@@ -611,7 +611,7 @@ function Router() {
       return "build";
     }
 
-    return "home";
+    return "builds";
   }, [path]);
 
   // Handle sidebar navigation
@@ -619,7 +619,7 @@ function Router() {
     (item: NavItem) => {
       const basePath = getEnvironmentPath();
       switch (item) {
-        case "home":
+        case "builds":
           navigateTo(basePath || "/");
           break;
         case "tasks":
@@ -722,7 +722,7 @@ function Router() {
           );
         }
         return (
-          <HomePage
+          <BuildsPage
             onNavigate={handleNavigation}
             onSelectBuild={handleSelectBuild}
             sidebarCollapsed={sidebarCollapsed}
@@ -730,10 +730,10 @@ function Router() {
           />
         );
 
-      case "home":
+      case "builds":
       default:
         return (
-          <HomePage
+          <BuildsPage
             onNavigate={handleNavigation}
             onSelectBuild={handleSelectBuild}
             sidebarCollapsed={sidebarCollapsed}
--- a/app/stardag-ui/src/api/tasks.ts
+++ b/app/stardag-ui/src/api/tasks.ts
@@ -51,6 +51,15 @@ export interface BuildFilters {
   status?: BuildStatus;
   // Only builds reactively scheduled by the named app.
   reactive_app_name?: string;
+  // Only builds with no activity for at least this long, measured on
+  // `last_activity_at`. Same definition and same 60s floor as
+  // POST /builds/bulk-cancel — deliberately the one predicate, so what a
+  // list shows and what a cleanup would act on cannot disagree. Ordering
+  // flips to stalest-first when it is set.
+  //
+  // Only `status=running` may accompany it (that pair is the
+  // abandoned-build query and stays exact); any other status is a 422.
+  idle_for_seconds?: number;
 }
 
 export async function fetchBuilds(
@@ -63,6 +72,8 @@ export async function fetchBuilds(
   if (filters.status) params.set("status", filters.status);
   if (filters.reactive_app_name)
     params.set("reactive_app_name", filters.reactive_app_name);
+  if (filters.idle_for_seconds)
+    params.set("idle_for_seconds", String(filters.idle_for_seconds));
 
   const url = `${API_BASE}/builds?${params.toString()}`;
   const response = await fetchWithAuth(url);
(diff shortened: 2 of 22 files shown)


andhus:
Copilot review triaged — both findings accepted, same class of bug.

1. **`fetchEpochRef` not bumped when the environment id is missing.** Real: an in-flight response can still pass `fresh()` and write state for the previous environment while the UI is gated. The epoch has to be invalidated on the bail-out path too.
2. **`BulkCancelDialog` dry-run not cancelled on unmount.** Real — closing the dialog quickly leaves a promise that resolves into `setState` on an unmounted component.

andhus:
**Now resolved** — both epochs.

`BuildsList` invalidates the in-flight request on both abandon paths (no active environment, and the env-change page reset). `BulkCancelDialog` gained an effect cleanup so closing it mid-dry-run cannot resolve into an unmounted component.

PR #216, opened 2026-08-07
feat(ui): explain stalled builds — cross-build blockers, claim holders, tick reasoning (#208 B5)

andhus:
Second frontend piece of **B5** in #208 — the "UI cannot explain a stalled
or spuriously failed build" half. Builds on the branch it targets.

## The incident this makes diagnosable

A reactive build failed within seconds of being triggered with:

```
No runnable or running tasks left but roots are not complete
(status counts: {'completed': 2, 'suspended': 1})
```

Nothing in the DAG was wrong. The real cause was a task left `RUNNING` by
a **different** build, holding an environment-global execution claim —
`latest_status` is denormalised per `(environment_id, task_id)`, so an
upstream any build leaves non-completed gates every downstream task in
every other build, while contributing nothing to *this* build's `running`
count. The UI showed a failed build, a task table with nothing obviously
wrong, and no way to find the task that was actually in the way.

Every fact needed to explain that is already in the API. This PR renders it.

## 1. "Why is this build not progressing?" on the build view

`GET /builds/{id}/frontier` has existed since reactive scheduling shipped
and the UI has never called it. It now drives a panel above the DAG.

**Show/hide rule** (`schedulingPanelForm`, unit-tested):

- **Prominent and expanded** when the build looks stalled: it has tasks,
  nothing actionable, and nothing running. Not a heuristic — that is
  *exactly* the condition under which the server computes
  `blocked_by_external`, and exactly the condition the SDK reads as "this
  build cannot progress", i.e. the read that produced the error above.
- Terminal states are qualified rather than blanket-included. Never for
  `completed`/`cancelled` (the build is done, or someone stopped it on
  purpose). For `failed`, only when the failure has the stuck-build
  shape: a blocker was reported, **or the build failed while nothing in
  it failed**. An ordinary DAG failure explains itself in the task table
  and does not need a scheduler post-mortem.
- **One quiet line** for a reactive build the rule does not flag, carrying
  the live counts and a disclosure onto the tick trail (which exists
  nowhere else in the UI).
- **Nothing at all** otherwise. A healthy non-reactive build has no
  scheduler to reason about, and a permanently empty box is worse than no
  box.

**The `blocked_by_external` ambiguity.** The server populates that list
only when the build has nothing actionable and nothing running, so an
empty list means *"not externally blocked, **or** not stalled"*. The panel
only ever says "no upstream outside the build is holding it back" from
inside the stalled branch, where the server did look. The progressing form
says the opposite explicitly — "upstreams owned by other builds are only
looked for while a build is stalled, so nothing here says whether this
build has any" — and there is a test asserting the progressing form makes
no no-blockers claim.

**Blockers.** Each entry names the blocked task of this build, the
blocking task (namespace-qualified — the viewer may never have seen it),
its status, how long it has been in it, and links to the build that put it
there. #208 calls that link "the single most useful piece of information
for the whole class of cross-build blocking problems". The copy splits on
`blocking_in_build`, because so does the remedy: a blocker inside this
build's own task set is at least visible in the table below, while one
outside it will never be scheduled here at all and can only be waited on.
`blocked_by_external_truncated` is stated rather than swallowed.

**Tick reasoning.** `GET /builds/{id}/tick-summaries`, newest first, two
deep with a "show N earlier ticks" disclosure. A stalled build repeats the
same outcome, so a trail reads as a diagnosis where one row reads as an
anecdote. Outcomes and known counters (`limit_denied`, `claim_denied`,
`self_healed`, …) get a label and a tooltip explaining what they imply;
**unknown keys render generically rather than being dropped**, because
`summary` is an open blob by design and the counter this UI has never
heard of is the one most likely to explain a new failure mode. A server
without the endpoint (404) degrades to "this server does not record tick
history", not an error.

## 2. Claim holder on the task detail panel

`StatusBadge` already hinted at a cross-build status with an icon. That is
enough to notice something, not enough to act. When a task's global status
is `RUNNING`/`SUSPENDED` under a build other than the one being viewed,
the panel now says "Execution claim held by another build", for how long,
with a link to the holder and the consequence spelled out.

Cancel logic is extended, not duplicated: the existing `window.confirm`
Cancel and the new dialog-confirmed release both go through one
`runTaskAction(action, targetBuildId)`. The plain Cancel button stands
down while another build holds the claim — it addresses the *viewed*
build, which is the wrong one, and two buttons that cancel the same task
under different builds is a trap rather than a choice.

## 3. Claim triage in the task explorer

A "Claims" mode beside the existing Search. The generic `/tasks/search`
filtering is untouched; this is a separate view because it needs a
different endpoint, and not for cosmetic reasons: `/tasks/search` reports
a task's status but neither *when* it entered it nor *which build* put it
there, and evaluates a status filter in Python over the whole unpaginated
match set. `/tasks?status=…&status_older_than=…` carries
`latest_status_at` and `latest_status_build_id`, filters in SQL, and
orders oldest-claim-first — the triage order.

Running/suspended toggles, a staleness cut (converted to an absolute
instant per request, since a duration would move the cutoff between
pages), a claim-holder column linking to the owning build, page-scoped
selection, and a bulk release that issues one cancel per task **under
that task's own owning build**. Partial failure is the normal case here
rather than an exception — a task can complete between the list being
drawn and the request landing — so the banner reports "released X of N"
and names each failure with the server's own reason instead of collapsing
into a single "failed". A task whose owning build was never recorded still
shows (it is a claim holder worth seeing) but is not selectable, because a
cancel has to be addressed to a build.

## Remedies and gating

`release` (cancel) for `RUNNING`/`SUSPENDED`, `retry` (reset to pending)
for `failed`/`cancelled`/`skipped`/`suspended` — mirroring what the server
will actually honour, since a button the API treats as a no-op is worse
than no button. Every one is addressed to the owning build and says which
in the confirmation. All are gated on `role === "owner" || "admin"`, the
same convention as concurrency-limit eviction and bulk build cancel;
non-admins keep the whole diagnosis and get a note instead of dead
buttons. Everything refetches after mutating.

## Notes

Data fetching follows the established idioms: no react-query, epoch-guarded
loads, and a refresh token from the build view so the 5s auto-refresh
drives one request stream rather than two. Tick history is fetched only
where it will be read. Every new surface carries explicit `dark:` pairs;
checkboxes and the repeated blocker action buttons carry accessible names
that identify their target.

## Tests

37 new tests (151 total, all green), covering: the show/hide rule across
running / progressing / no-tasks / completed / cancelled / failed-with-a-
failure / failed-with-none; an external blocker rendering with a working
owning-build link; `blocking_in_build` true vs false copy; the empty-list-
while-progressing case making no no-blockers claim; truncation; an unknown
tick-summary counter key surviving; a 404 from tick-summaries degrading
gracefully; a frontier read failure surfacing; release/retry confirming and
addressing the owning build then refetching; admin gating on all three
surfaces; the claim query's statuses and absolute staleness cutoff; a
non-selectable row with no recorded owning build; and bulk release
reporting partial failure per task.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,46 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
   that every replica runs its own timer with no leader election; cancellation
   is idempotent, so concurrent sweeps are wasteful rather than wrong.
 
+### UI
+
+- **A Builds view, replacing "Home".** Builds are listed with status,
+  duration, last activity and the reactive app that owns them, and can be
+  filtered by status, by owning app, and by how long they have been idle.
+  The idle filter means _abandoned_, so it implies Running: a finished
+  build has no activity by definition, and including terminal builds would
+  fill a staleness listing with history sorted oldest-first.
+
+- **Bulk cleanup of abandoned builds**, admin-gated to match the API. Rows
+  are selectable per page, and "Clean up idle builds…" sweeps the whole
+  environment on the server rather than the visible page. Both paths open a
+  confirmation showing a **dry run of the real query** — the same selection
+  `POST /builds/bulk-cancel` will act on, including the per-build reasons
+  anything was skipped — because a preview that disagrees with the action
+  is worse than no preview.
+
+- **A build that is not progressing now says why.** When nothing is
+  actionable and nothing is running, the build view names each blocking
+  upstream, its status, how long it has been held, and which build owns it
+  — including blockers that are not part of the build at all, which are
+  otherwise invisible from it. Each carries the remedy (release the claim,
+  or reset the task to pending) addressed to the owning build. Behind a
+  disclosure: the scheduler's own tick trail, with runs of identical ticks
+  collapsed ("lease held ×20") and counters that stayed at zero dropped, so
+  a stalled build's repetition reads as a diagnosis instead of a log.
+
+- **Claim triage in the task explorer.** Lists the tasks holding an
+  execution claim, longest-held first, with the build that owns each one,
+  and releases them in bulk. Only tasks actually holding a claim are
+  selectable, and every release reads the resulting status back — a task
+  that finished between listing and acting is reported as such rather than
+  counted as released.
+
+- **Fixed:** unchecked checkboxes rendered as white boxes in dark mode
+  (`color-scheme` was never set, so every native control used the light
+  theme); durations past a day read as `371h 41m` instead of `15d 11h 41m`;
+  dates rendered in the browser's locale order (`8/9/2026` or `9/8/2026`
+  depending on the reader) rather than `YYYY-MM-DD`.
+
 ### SDK
 
 - **Reactive builds now discover the DAG inside Modal, not on the machine
--- a/app/stardag-ui/src/api/tasks.ts
+++ b/app/stardag-ui/src/api/tasks.ts
@@ -1,8 +1,10 @@
 import type {
   Build,
   BuildCancelResult,
+  BuildFrontier,
   BuildListResponse,
   BuildStatus,
+  BuildTickSummaryListResponse,
   BulkCancelBuildsRequest,
   BulkCancelBuildsResponse,
   Task,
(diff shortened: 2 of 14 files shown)


andhus:
Copilot review triaged — one real, one a misreading that is still worth a doc change.

**`wantTicks` comment.** Real, and my fault: the comment says tick history is fetched "always for a stalled build", but a later change put it behind the details disclosure. The comment is now wrong.

**`schedulingPanelForm` returning `collapsed` for terminal reactive builds.** Not a bug — that is intended and is asserted by the existing tests. The docstring's "completed / cancelled: never" refers to the **stalled** form, not to the quiet one-line strip. But if it read that way to a reviewer it is under-specified, so the docstring should say which form it is talking about.

andhus:
**Now resolved** — both comments corrected.

`wantTicks` no longer claims tick history is fetched "always for a stalled build" (the compaction pass put it behind the disclosure). `schedulingPanelForm`'s terminal-state paragraph now says explicitly that it governs the `"stalled"` form only — a terminal reactive build still gets the quiet strip, which is the sole route to its tick trail.

PR #217, opened 2026-08-07
feat(cli): stardag builds/tasks groups, and report tick summaries to the registry (#208 B4/B6)

andhus:
Closes the operational half of #208: **B4** (there is no `stardag builds`
CLI surface at all) and **B6** (tick outcomes are not persisted).

B4 opens with why this matters — the incidents in that issue were hard to
diagnose because the failure was visible in the executor's logs and in the
UI, but *"what does the scheduler actually think the state is?"* required
hand-rolled API calls. `builds frontier` in particular would have made the
A1 bug self-diagnosing.

Scope is `lib/` and `docs/` only. The server endpoints this drives landed
in the sibling API PRs.

---

## The commands, and what each answers

```
stardag builds list [--status running] [--reactive-app NAME] [--older-than 24h]
                    [--page N] [--limit N] [--json]
stardag builds show <build-id> [--json]
stardag builds frontier <build-id> [--json]
stardag builds ticks <build-id> [--limit N] [--json]
stardag builds cancel <build-id> [--cascade/--no-cascade] [--yes]
stardag builds cleanup [--older-than 24h] [--build-id ID ...] [--reactive-app NAME]
                       [--include-reactive] [--cascade/--no-cascade] [--limit N]
                       [--reason TEXT] [--apply] [--yes] [--json]

stardag tasks list [--status running ...] [--older-than 1h] [--name N]
                   [--namespace NS] [--page N] [--limit N] [--json]
stardag tasks cancel <build-id> <task-id> [--yes]
stardag tasks retry <build-id> <task-id> [--yes]
```

All accept `-p/--stardag-profile` and `-e/--stardag-env`, like
`concurrency-limits`.

| command | question it answers |
| --- | --- |
| `builds list` | which builds exist, what state are they in, and how long have they been idle |
| `builds show` | one build: status, roots, reactive metadata, and its two liveness timestamps spelled apart |
| `builds frontier` | what a scheduler tick sees — and, when the build looks stalled, *who is holding it back* |
| `builds ticks` | what each recent tick decided, and why (B6's data) |
| `builds cancel` | stop a build; `--cascade` also releases the claims its tasks hold |
| `builds cleanup` | find and terminate abandoned builds, releasing their claims |
| `tasks list --status running` | which tasks hold an execution claim, since when, and under which build |
| `tasks cancel` | release a claim stranded by a build that is gone |
| `tasks retry` | reset a failed/cancelled/skipped/**suspended** task to PENDING |

Two adaptations to the issue's proposal, called out as asked: `builds ticks`
is new (B6's summaries are only useful if something reads them, and this is
the "why is this build not progressing?" answer), and the destructive
commands take the `--yes` confirmation gate the existing CLI already uses
for `concurrency-limits delete/evict`.

## `builds frontier` — the diagnostic command

Besides the actionable/running partitions it renders the build's external
blockers: which task of this build is held up, by which task (namespace and
name, not just an id — the blocker may be entirely unknown to this build, so
an id alone diagnoses nothing), in what status, for how long, and which build
owns it. It distinguishes `blocking_in_build` true from false, because the
remedies differ:

- **not in this build's task set** — this build will never schedule it and can
  only wait for the owner; if the owner is gone, release the claim with
  `stardag tasks cancel <owning-build-id> <task-id>`;
- **in this build's task set** — it resolves when the owning build finishes
  it; a retry from here would not release the claim.

It respects `blocked_by_external_truncated`, and it handles the field's
semantics honestly. The server populates the list **only when a build has
nothing actionable and nothing running**, so an empty list means "not
externally blocked *or* not stalled". For a build that is merely progressing
the output says the list was **not evaluated**, naming the counts that prove
it; only for a build that really has nothing to do does it report that no
blockers were found — and even then it names the other possibility (a server
predating the fields). Asserting "no blockers" for a healthy build would be a
new version of exactly the misdirection this command exists to end.

## `builds cleanup` — the B3 runbook

Defaults to a **dry run**: it prints what a real run would cancel — the
builds, the claims that would be released, the per-build `skipped` map with
its reasons, and the truncation flag — and writes nothing.

**`--apply` is the only thing that makes it act.** `-y/--yes` only skips the
confirmation prompt, so `cleanup -y` on its own is still a dry run: on a
command that reports by default, `-y` alone becoming a cascade of
cancellations is exactly the surprise a destructive command may not have.
`--apply` prompts; `--apply --yes` runs unattended.

The selection is the **server's** in both passes — same filter, same
endpoint, `dry_run` only flipped — rather than reimplemented client-side, so
what you review is necessarily what you get.

```sh
stardag builds list --status running --older-than 24h   # find them
stardag builds show <build-id>                          # inspect one
stardag builds frontier <build-id>                      # what is it holding / held by
stardag tasks list --status running --older-than 24h    # or ask claim-first
stardag builds cleanup --older-than 24h                 # dry run (default)
stardag builds cleanup --older-than 24h --apply         # act (prompts)
stardag builds cleanup --older-than 24h --apply --yes   # act unattended
```

## The `--json` convention

The CLI had **no** machine-readable output convention (`--json` appeared only
as an *input* option in `environment.py`), so this establishes one:

- `--json` is on the read-only commands — `builds list/show/frontier/ticks`,
  `tasks list` — and on `builds cleanup`, whose default dry run is read-only
  and whose payload is the same shape either way.
- **stdout carries exactly one JSON document and nothing else.** Every hint,
  warning, note and prompt goes to stderr, so `| jq` is always safe. Emitted
  with `typer.echo`, not the rich console, which would syntax-highlight and
  soft-wrap.
- The document is the SDK's model of the API payload —
  `model_dump(mode="json")`: the same field names and nesting as the REST
  response, minus any field this SDK version does not model. Machine-readable
  codes stay raw (e.g. `skipped` keeps `not_running`, not its expansion).
- The page payload is the server's, verbatim — nothing is filtered or
  re-sorted client-side, so `total` and the rows always agree with what the
  server was asked for.
- `builds cleanup --json --apply` requires `--yes` rather than prompting into
  the pipe.

## Durations

One integer, one optional unit of `s`/`m`/`h`/`d`/`w`; a bare number is
seconds. `24h`, `90m`, `3d`, `2w`. Compound forms (`1h30m`) and fractions
(`1.5h`) are rejected rather than silently interpreted, months/years are not
units, and **zero is rejected** because `--older-than 0` reads as
"everything" and a staleness filter that matches live work is the one mistake
this surface exists to prevent. Build staleness additionally enforces the
server's 60s floor client-side, so the grammar error lands in front of the
user rather than as a 422 from three layers down.

The registry deliberately speaks two dialects of "how old" —
`idle_for_seconds` for builds (a recurring *policy*) and an absolute
`status_older_than` for task staleness (so a paged scan's cutoff can't
drift). One parser, converted at the boundary to whichever the endpoint
takes.

**The filter is applied server-side**, by the same SQL predicate the reaper
and `bulk-cancel` use — so `builds list --older-than 24h` and
`builds cleanup --older-than 24h` agree on what is stale by construction,
`total` stays an exact count, pagination stays server-side, and the server
orders stalest-first. On `builds list` it may only be combined with
`--status running`; only RUNNING has a SQL predicate, and the CLI rejects the
other combinations itself rather than surfacing the server's 422 raw.

A CLI can be newer than the registry it talks to, and FastAPI ignores an
unknown query param silently rather than rejecting it — so `--older-than`
against an old server would quietly mean "everything". The command detects
that (a returned build newer than the cutoff, or one with no
`last_activity_at` at all — a server that cannot report the field cannot have
filtered on it) and prints a loud **stderr warning** that the registry does
not support the filter and the results are unfiltered. It deliberately does
*not* filter the page itself: the server already paginated and counted
without the filter, so a local cut would drop rows from a page chosen wrong,
under-reporting exactly the oldest builds. Warn, never silently truncate.

## SDK registry client

New on `RegistryABC` (with safe defaults) and `APIRegistry`:

```python
build_list(*, page, page_size, status, reactive_app_name, idle_for_seconds) -> BuildListPage
build_get_summary(build_id) -> BuildSummary
build_cancel(build_id, *, cascade=False) -> BuildCancelResult | None          # + _aio
build_bulk_cancel(*, build_ids, idle_for_seconds, reactive_app_name,
                  include_reactive, cascade, dry_run, limit, reason) -> BulkCancelResult
build_report_tick_summary(build_id, summary)                                  # + _aio
build_list_tick_summaries(build_id, limit=20) -> list[TickSummaryRecord]
task_list(*, page, page_size, status, status_older_than,
          task_name, task_namespace) -> TaskListPage
task_cancel_by_id(build_id, task_id)                                          # + _aio
task_retry_by_id(build_id, task_id)                                           # + _aio
```

Notes:

- **`build_cancel` returning `BuildCancelResult | None`** follows
  `task_register_bulk`'s optional-return convention — lifecycle callers
  ignore it, operator tooling reads the cascade. No new method for one
  endpoint.
- **`build_list_running` is now expressed in terms of `build_list`** with
  `status="running"`, i.e. filtered server-side. It used to page an
  unfiltered list and match the derived status in Python, so an environment
  holding more non-running builds than the sweep's page budget could starve
  the watchdog of the running builds it exists to find — silently disabling
  the safety net exactly when a backlog makes it necessary (A3). Ordering,
  page budget and truncation warning unchanged.
- **`build_get_summary` is a second read model over the same endpoint as
  `build_get`, deliberately.** `BuildInfo` is the contract a *custom* registry
  backend must satisfy for reactive scheduling — four fields it necessarily
  has. `BuildSummary` is the operator view of an API-registry build: names,
  timestamps, liveness. Fusing them would make every reactive-capable backend
  responsible for fields it has no notion of.
- **`task_cancel_by_id` / `task_retry_by_id`** exist because operator tooling
  only ever has the id, and rehydrating a task object to cancel it would fail
  for exactly the abandoned tasks that most need cancelling. The task-object
  variants now delegate to them.
- `task_list` builds its params as a list of pairs, because `?status=` is
  repeatable and a mapping cannot express two values for one key.
- New models (`BuildSummary`, `BuildListPage`, `BuildCancelResult`,
  `BulkCancelResult`, `BulkCancelBuildRef`, `TaskSummary`, `TaskListPage`,
  `TickSummaryRecord`) ignore unknown response fields and default every field
  a server might not send — forward *and* backward compatible.
  `BuildSummary` documents at length why `last_active_at` and
  `last_activity_at` are two different numbers, since measuring staleness on
  the first would call a build that has been running tasks for three days
  idle.

## B6 — tick summaries

`TickSummary` is rich and went only to the container's log. A reactive build
is driven by dozens of short-lived ticks, so reconstructing why one stalled
meant reading logs across all of them, and the counters that explain a stall
were computed and thrown away.

Guarantees:

- **Best-effort is a contract, not a hope.** This runs at the end of every
  tick; a reporting failure can never fail a tick or change its outcome.
  Bare `except Exception` — the correct breadth precisely because nothing this
  call can raise is worth propagating to a scheduler — logged at warning;
  version-skew 404s at debug.
- **Missing route tolerated** the way `_skip_blocked` does it
  (`is_missing_route_error`), and **latched process-wide** so an SDK pointed
  at an older server stops paying for a doomed request every tick. A
  *resource*-level 404 (the build is gone) does not latch — it is not a reason
  to stop reporting for every other build in the process.
- **One call site.** `run_tick_aio` had five returns across a nested loop;
  rather than a report at each — where a return added later would silently
  stop being observable — the scheduling moved to `_run_tick_body_aio`, which
  mutates the summary in place, and the wrapper reports once on the way out.
- **A tick that raises is reported too**, as `outcome="error"` carrying
  `TickSummary.error_type` and `error_message` — the most informative answer
  a "why did this build stall?" query can get, and the open-blob storage
  takes the new fields without a server change. Same discipline as the
  trigger's terminal-event wrapper: report best-effort, **never mask or
  replace the original exception**, re-raise unconditionally, and swallow
  (with a log) any failure to record the failure. The message is clipped to
  1024 chars with an explicit marker, so a traceback-sized string cannot push
  the summary past the server's 8 KiB compact-JSON cap and turn a recorded
  failure into no record at all.
- **Which outcomes:** everything except `not_reactive` — excluded by
  definition (the build is not reactively scheduled, so the tick did nothing
  and learnt nothing) and retention is finite. `lease_held` *is* reported:
  contention is signal.
- **`TickConfig.report_tick_summaries`** (default `True`) disables it, and is
  deliberately **not** in `_TICK_KWARGS_ALLOWED`. Following
  `stale_running_no_ref_seconds` and `stale_external_blocker_seconds`: the
  reasons to change it — a registry without the endpoint, a deployment that
  keeps observability elsewhere — are properties of the whole deployment,
  never of one build. Widening that allowlist later is additive; narrowing it
  is not.
- The summary is stored verbatim server-side apart from a required `outcome`,
  so future `TickSummary` fields need no server release; a test pins that the
  whole dataclass rides.

## Refactor called out

`_resolve_registry` / `_fail` / the profile+env options were private to
`_cli/limits.py`. Three groups needing the same 80 lines is three copies that
drift, so they moved to `_cli/_registry_ctx.py` and are imported back into
each group's namespace — which keeps the established test seam working, since
patching `stardag._cli.<module>._resolve_registry` still intercepts the
lookup. `limits.py` behaves identically; only the patch targets for the
resolver's *internals* moved, so those tests moved with them into
`tests/test__cli/test_registry_ctx.py`.

## Tests

`1171 passed, 29 deselected` (was 1053 on the base) — `uv run pytest tests -q
-m "not modal_live"`; `uv run pyright src tests` clean; `pre-commit` clean on
every changed file.

- `tests/test__cli/test_duration.py` — the grammar, including every rejection
  (compound, fractional, bare unit, internal whitespace, non-fixed units,
  zero) and the coarse renderer.
- `tests/test__cli/test_builds.py` (40) — every command; `--older-than`
  conversion and its rejections (bad grammar, below the 60s floor, paired
  with a non-`running` status), the server-ignored-it warning **and** that
  the rows survive unfiltered, the no-timestamp case, and the no-warning
  case; `--json` parseable *and* uncontaminated (including when a stderr
  warning is emitted, where the payload stays the server's page verbatim);
  the dry-run-then-apply flow, `-y` alone still being a dry run,
  `--apply --yes` running with no input available, and `--json --apply`
  refusing to prompt; blocker rendering with identity/age/owner, the
  truncated case, the in-build vs out-of-build remedies, and both empty-list
  cases (progressing vs genuinely stalled); empty-state hints; error paths
  closing the client.
- `tests/test__cli/test_tasks.py` (14) — repeatable `--status`, the absolute
  cutoff, claim-holder rendering, `--json`, confirmation gates, error paths.
- `tests/test__cli/test_registry_ctx.py` — the moved resolver tests.
- `tests/test_registry/test_api_registry.py` — wire-level via
  `httpx.MockTransport`: filters as query params, unset filters omitted,
  unknown response fields ignored, missing liveness fields defaulting,
  `status=running` reaching the server for the watchdog sweep, `cascade`,
  bulk-cancel body/parsing, repeated `status=`, ISO-8601 `status_older_than`,
  id-addressed paths, and tick-summary POST body verbatim / GET parsing.
- `tests/test_build/test_reactive.py` — reported on `terminal`,
  `lingered_out` and `lease_held`; skipped for `not_reactive`; disabled by
  config; a raising registry leaving the tick's result untouched; a missing
  route tolerated *and* latched; a resource 404 not latching. Plus the error
  outcome: a crashed tick recorded with type/message while the exception
  propagates unchanged, the message bounded (asserted against the 8 KiB
  cap), a dead reporter not masking the original exception, and the config
  toggle covering the crash path too.

## Docs

`docs/docs/configuration/cli.md` gains a **Build & Task Commands** section
with the commands, the duration grammar, the `--json` convention, **Reading
the frontier** and **Cleaning up abandoned builds** (the end-to-end runbook,
with the four things easy to get wrong stated explicitly). Cross-linked from
the cross-build-blocking concept section and the Modal how-to, which also
gains the tick-reporting guarantees. `CHANGELOG.md` under `[Unreleased]`.

## It removes the `reactive_app_name` signature shim

Called out because it is easy to read as an accident in the diff. #209
added `accepts_reactive_app_name_kwarg` so the watchdog could degrade to an
unscoped listing against a `RegistryABC` implementation predating the
kwarg. This PR re-expresses `build_list_running` in terms of `build_list`
with `status="running"`, which makes the scoping **server-side for every
registry** — so the shim has nothing left to do, and it goes, along with
the two conditional call sites and their tests.

Same direction as #219, which drops the remaining compatibility shims for
registry backends that do not exist. Noted here rather than deferred to
#219 because the shim's whole reason for existing disappears in *this*
change.

---

Stacked on #214 (`feat/reactive-external-blocker-handling`) — review that
first.


andhus:
Copilot review addressed — two findings fixed, two were reading a deliberate change as accidental.

**`--reactive-app` not forcing `--include-reactive`.** Real, and worse than a help-text mismatch: the server excludes reactive builds by default, so naming an app selected exactly the population the same request then filtered out, reporting no matches for a filter that matched. Fixed at the call site.

**`build_list` docstring.** Real. It claimed the CLI re-applies the idle cut locally against an older server. It does not, deliberately — `total` and the page boundaries are computed server-side over the unfiltered set, so trimming the page locally gives a wrong count and paginates over the wrong population. It warns instead. Docstring now says that.

**The two `reactive_app_name` compatibility comments** (`_base.py` and `_app.py`) describe an intentional change, not a regression. This PR re-expresses `build_list_running` in terms of `build_list`, which makes the scoping server-side for *every* `RegistryABC` — so the signature-inspection shim added in #209 has nothing left to do and is removed here along with `accepts_reactive_app_name_kwarg`. That is the same direction as #219. Worth calling out in the PR description, which I have not yet updated.

andhus:
**Now resolved** — plus a description correction.

- `builds cleanup --reactive-app X` now forces `include_reactive=True`, so naming an app no longer selects the population the same request then excludes.
- `build_list`'s docstring no longer claims the CLI re-applies the idle cut locally; it warns instead, because trimming a page locally yields a wrong `total` and paginates over the wrong population.
- `--older-than` docs and CHANGELOG now say it **implies** `--status running` rather than merely 'can only be combined with' it.
- The PR description now calls out that this PR removes the `accepts_reactive_app_name_kwarg` shim, which was the thing the two remaining review comments were pointing at. It is deliberate: expressing `build_list_running` via `build_list` makes the scoping server-side for every registry, so the shim has nothing left to do.

Verified: SDK suite 1257 passed, and the merged stack redeployed to a live Modal+Neon environment.

PR #218, opened 2026-08-07
docs(design): note on execution claims and liveness

andhus:
Adds `docs/design/` — maintainer-facing notes on *why* parts of stardag are built the way they are. Deliberately outside the published site (`docs/docs/`): user docs describe how to use stardag, these describe trade-offs a contributor needs before changing something, including things stardag does badly and knows it.

A note earns a place there when a design keeps getting re-litigated, or keeps being misread the same way.

## The first note: execution claims and liveness

How "is this task being executed right now?" is modelled, and why it is a **status** rather than a lease.

**The advantages are load-bearing, not incidental**, and the note says so plainly so nobody undoes them casually:

- one row, one transaction — claim, status, completion and concurrency-slot occupancy are written together and *cannot* disagree; a lock table would immediately create a reconciliation problem
- sticky `COMPLETED` composes for free rather than becoming a cross-entity rule
- every reader is one indexed column, no join — on a frontier that is re-read every few seconds per active build
- **zero liveness traffic**: a real lease needs heartbeats (~360 extra writes per six-hour task; ~6 writes/second sustained at a thousand concurrent tasks) purely to say "still alive"
- fewer user-facing concepts

**The cost is narrower than "the claim has no TTL."** Within a build the claim carries an executor ref, and probing the backend is *better* evidence than any heartbeat. Only across builds is there nothing to evaluate — build X cannot probe build Y's execution. So the actual gap is:

> The claim records no liveness evidence that a **third party** can evaluate.

The note tabulates the machinery that exists to infer that from outside, sketches the one-column fix it implies (`latest_status_expires_at`, written once, no heartbeats, `NULL` = today's behaviour), and is explicit about what that fix would **not** remove — executor probing, build-level reaping, the two-phase start.

## Half of it is corrections

These were each misread more than once during #208, which is the whole justification for writing them down:

1. **"It needs a lease" overstates the case.** A lease doesn't replace probing; `stale_running_no_ref_seconds` guards a seconds-long crash window, not a core mechanism; build-level reaping survives regardless.
2. **Identity and type resolution are *not* welded together.** Hash-mode serialization is separate by design (`CONTEXT_MODE_KEY`, as `HashableSet` uses) — a load-bearing piece of the stardag↔pydantic integration, not a workaround. Regular-mode serialization is therefore freely extensible; the hash payload is the constrained one, on purpose. Hashing namespace and task family is a deliberate feature for controlling identity collisions.
3. **A `__module` hint in the payload is not the cheap alternative it looks like.** It does not survive refactoring (a stored module path records where a class *was*), and importing a module name taken from registry data is an execution primitive needing an allowlist — that is, a declaration, which is `task_modules` again but implicit and unauditable.
4. **`last_active_at` and `last_activity_at` are different things.** The former moves only on lifecycle transitions, so reaping or displaying it calls a busy build idle.
5. **`blocked_by_external` is only computed when a build looks stalled**, so an empty list means "not externally blocked *or* not stalled" — never proof of no blockers.

Docs-only; no code changes.

### DIFF
--- a/docs/design/README.md
+++ b/docs/design/README.md
@@ -1,0 +1,18 @@
+# Design notes
+
+Maintainer-facing notes on _why_ parts of stardag are built the way they
+are — the constraints in play, the alternatives that were rejected, and the
+known costs of what was chosen.
+
+These are deliberately **not** part of the published documentation site
+(`docs/docs/`, see `mkdocs.yml`). User docs describe how to use stardag; these
+describe trade-offs a contributor needs before changing something, including
+things stardag does _badly_ and knows it.
+
+A note earns its place here when a design keeps getting re-litigated, or keeps
+being misread in the same way. If you find yourself explaining the same
+subtlety a second time, write it down here instead.
+
+| Note                                                                 | Subject                                                                                       |
+| -------------------------------------------------------------------- | --------------------------------------------------------------------------------------------- |
+| [execution-claims-and-liveness.md](execution-claims-and-liveness.md) | How "is this task running right now?" is modelled, and why it is a status rather than a lease |
--- a/docs/design/execution-claims-and-liveness.md
+++ b/docs/design/execution-claims-and-liveness.md
@@ -89,16 +89,25 @@ So the gap is narrower and more specific than "no TTL":
 > The claim records no liveness evidence that a **third party** can evaluate.
 
 Everything below exists to infer, from the outside, what a single stored
-expiry would state directly:
+expiry would state directly.
+
+Some of these landed with the reactive-scheduling work this note was
+written alongside, so a reader on `main` will not find all of them yet —
+they are marked. The argument does not depend on which are merged: each is
+an independent attempt to answer "is that claim still alive?" without the
+claim saying so, which is the point.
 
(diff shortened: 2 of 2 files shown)


andhus:
Copilot review triaged.

**References to things not in-tree.** Fair for a reader of this branch: the note is based on `main`, and `TickConfig.stale_external_blocker_seconds`, `blocked_by_external`, `services/build_cleanup.py` and the task-status listing all arrive in the accompanying PRs (#210, #213, #214). The note presents them as existing machinery, which is true of the stack and not of `main`.

The fix is framing, not deletion — the note should say which mechanisms are introduced by the accompanying work rather than implying they are already there.

andhus:
**Now resolved.**

The mechanisms table marks with (†) the entries introduced by the accompanying reactive-scheduling work, with a note saying a reader on `main` will not find them yet. Marked rather than removed: the argument is that each is an independent attempt to answer "is that claim still alive?" without the claim saying so, and that holds whichever are merged.

The "(not implemented)" heading is now historical too — the expiry landed close to the sketch (#220). Said so, and kept the sketch as written, since a note rewritten to match its outcome stops showing how the decision was made.

PR #219, opened 2026-08-07
refactor(sdk): drop compatibility shims for registry backends that do not exist

andhus:
Pure deletion PR: removes compatibility machinery kept alive for
`RegistryABC` subclasses written against older method signatures. Every
item below was verified by searching the repo first — the only
implementations are `APIRegistry`, `NoOpRegistry` and test fakes, all of
which take the current signatures.

**Net: 252 insertions, 671 deletions (−419); source only: 125 / 314 (−189).**

Behaviour is unchanged for `APIRegistry` and `NoOpRegistry` against a
current server. Note what is *not* touched: SDK↔server **version skew**
(the API deploys separately from the SDK) is real and stays — missing-route
handling, fields absent on older servers, `limit_key`/`claim` params a
older server ignores. Tolerating an older *server* is a different thing
from tolerating an older *subclass*.

### 1. Signature-reflection helpers (`registry/_base.py`)

`accepts_executor_kwargs`, `accepts_executor_metadata_kwarg` and the
cached `_param_info` behind them inspected a registry method's signature
at every call site so the SDK could drop `executor` / `executor_ref` /
`executor_metadata` for subclasses that predated them. 13 call sites
across `registry/_base.py`, `build/_reactive.py`, `build/_concurrent.py`
and `integration/modal/_app.py`, each doubling a code path. Deleted, and
every caller collapsed to the single call it should have been.

Neither helper was exported from `stardag.registry`; callers reached into
`stardag.registry._base`. `executor_metadata=None` is still omitted from
the wire by `APIRegistry._get_start_params`, so what the server sees is
unchanged.

(`accepts_reactive_app_name_kwarg` no longer exists on this base — nothing
to delete.)

### 2. `RegistryABC.task_start_claim_aio`'s no-arbitration default

It recorded a plain start and reported "you won". For a method whose
entire purpose is exactly-once arbitration, a backend that silently grants
every claim is indistinguishable from one that arbitrates correctly.

It now raises `NotImplementedError`. **`NoOpRegistry` overrides it with an
explicit unconditional grant** — that is the registry-*less* path: no
shared state records that a task is RUNNING, so there is nothing to
arbitrate against and no cross-build guarantee on offer either way.
Granting is the honest answer there, and it keeps the registry-less path
working.

Deliberately *not* `@abc.abstractmethod`: `RegistryABC` requires only
`task_register` / `task_get_metadata` today, and subclassing
`NoOpRegistry` is the documented way to opt out. Documented in the
docstrings and in `docs/docs/concepts/build-execution.md`.

### 3. The `claim_active` / `_registry_supports_claim` gates

Both engines carried
`type(registry).task_start_claim_aio is not RegistryABC.task_start_claim_aio`
to avoid calling the default. Gone from `build/_reactive.py` and
`build/_concurrent.py` (along with the "does not implement
task_start_claim_aio — proceeding without claims" warning). For
`APIRegistry` the gate was already always True, so nothing changes.

### 4. `TickConfig.claim`

A switch to turn exactly-once off. Removed, together with the
non-claiming limits-only acquisition branch it guarded in
`_act_on_frontier` — which collapses that function from a four-way branch
to a single claiming start carrying `limit_keys`.

### 5. The duplicate missing-route alias

`registry/_api_registry._is_route_not_found` just called
`exceptions.is_missing_route_error`, and `build/_reactive.py` imported
both — reaching into `_api_registry` for a check that lives in
`exceptions`. One name now. This is the version-skew case, so the
behaviour is untouched.

### `RegistryABC.task_start_with_limits_aio` — kept

The premise ("strictly subsumed by `task_start_claim_aio(limit_keys=...)`")
holds for the reactive tick, and its use there is deleted. But
`RegistryConcurrencyLimiter` (`build/_registry_limiter.py`) needs a limit
acquire that does **not** arbitrate, and routing it through the claiming
method would deadlock it:

- The resident engine takes the claim **before** the limiter slot (the
  registry-backed limiter counts RUNNING tasks, so claiming inside the
  slot would deny itself — already documented in
  `docs/docs/concepts/build-execution.md`).
- So by the time the limiter acquires, the task is already RUNNING because
  of the engine's own claim. A claiming acquire there returns
  `denied_reason="already_running"`, which `_acquire` reads as "at
  capacity" and retries — forever.

Two SDK methods over one endpoint, but they are two different semantics
(arbitrating vs not), not two vintages. `_act_on_frontier`'s four-way
branch — the stated motivation — is collapsed either way.

### `emit_skips_for_blocked_tasks` (`build/_concurrent.py`) — kept

It is not a pure duplicate of `POST /builds/{id}/skip-blocked`:

- The resident engine's default registry **is** `NoOpRegistry`
  (`init_registry()` returns it whenever no registry is configured), and
  `RegistryABC.build_skip_blocked` defaults to returning `[]`. On the
  registry-less path the endpoint yields nothing, so the local fixpoint is
  the only thing that works.
- Even with a server, the endpoint returns task ids only. The local loop
  also sets `state.exception = _SkippedSentinel()` (which
  `find_ready_tasks` and the deadlock/blocked detection read) and
  `task_count.skipped`, which is returned to the caller in
  `BuildSummary.task_count`. Neither is derivable from the response.

`tests/test_build/test_concurrent.py::test_skips_emitted_for_transitively_blocked_tasks_continue`
asserts exactly this on a `NoOpRegistry` subclass.

## Test coverage

Tests that existed only to exercise a deleted fallback are deleted, not
left asserting a path that no longer exists:

| Deleted / rewritten | What still covers the surviving path |
| --- | --- |
| `TestOldRegistrySignatureCompat` (3 tests), `TestExecutorMetadataOnStart::test_metadata_dropped_for_pre_metadata_registry` (test_detached.py) | `test_handle_metadata_recorded_on_start` |
| `TestAcceptsExecutorMetadataKwarg`, `LegacyRegistry` + its two forwarding cases (test_executor_metadata.py) | `TestRegistryABCForwarding` (rewritten to forwards-only), `TestAPIRegistryWireFormat` |
| `test_metadata_dropped_for_pre_metadata_registry`, `test_acquiring_start_metadata_dropped_for_legacy_registry` (test_reactive.py) | `TestExecutorMetadataRecording`, `TestAcquiringStartExecutorMetadata::test_acquiring_start_carries_metadata` |
| `test_started_drops_metadata_for_legacy_registry` (modal test_executor_metadata.py) | `test_started_passes_metadata_to_aware_registry` |
| `test_claim_off_uses_legacy_limits_path` (deleted with `TickConfig.claim`) | — |
| `test_claim_skipped_without_registry_support` → `test_claim_through_registry_less_path` | asserts `NoOpRegistry` grants and the build succeeds |
| `test_no_selector_no_limit_calls` → `test_no_selector_claims_without_limit_keys` | asserts the claim carries no limit keys |

`FakeReactiveRegistry` now implements real claim arbitration mirroring the
API's claim-on-start — the tick always claims, so the base fake has to.
`ClaimingReactiveRegistry` keeps only its scriptable cross-build race.
New `tests/test_registry/test_start_claim.py` covers both halves of the
new contract (`RegistryABC` refuses, `NoOpRegistry` grants).

## Verification

```
$ uv run pytest tests -q -m "not modal_live"
1160 passed, 29 deselected, 15 warnings in 19.66s

$ uv run pyright src tests
0 errors, 0 warnings, 0 informations
```

(1171 → 1160: 15 fallback-only tests deleted, 4 added.) `pre-commit` clean
on every changed file. The `modal_live` tier was not run.


andhus:
Copilot review triaged.

**Test-name typo (`defaultis`).** Accepted — cosmetic, and worth fixing since it shows up in test output.

andhus:
**Now resolved** — `defaultis` → `default_is`.

PR #220, opened 2026-08-07
feat(api): give the execution claim an expiry so abandoned claims self-heal

andhus:
`Task.latest_status == RUNNING` **is** the execution claim — arbitrated in `_create_task_event` under `SELECT … FOR UPDATE`, in the same transaction as the event and the concurrency-limit slot rows. That single-row design stays exactly as it is, and its advantages are load-bearing: one transaction so claim/status/completion/slots cannot drift, one indexed column with no join on a frontier re-read every few seconds, and zero liveness traffic. This is **not** a lock table and **not** a lease; nothing heartbeats.

Its one defect is that the claim recorded no liveness evidence a **third party** could evaluate. Within a build the claim carries an executor ref that can be probed — better evidence than any heartbeat, and probing stays. Across builds there was nothing at all, so a holder that vanished wedged the task for every future build forever and leaked its concurrency-limit slots with it.

This PR adds one nullable column and the predicates that honour it. An expired claim is simply re-claimable — no reaper, no release call, no new status a user has to understand.

## The contract (an SDK PR consumes these verbatim)

**Column** — `tasks.latest_status_expires_at` (`timestamptz`, nullable). `NULL` = "no expiry known", and every predicate treats it as a claim that never lapses — exactly today's behaviour, so the column is additive by construction.

**The migration backfills every row that is RUNNING when it lands**, as `latest_status_at + <default TTL>`. Without that the feature would ship not fixing the cases that motivated it: those rows *are* the abandoned claims — a task RUNNING since three months ago is not "never lapses" — and a consumer that correctly reads `NULL` as "wait, unbounded" would wedge on them forever. The value is right in both directions without the migration having to tell them apart:

- abandoned long ago → the value is already in the past → immediately re-claimable, which is the heal;
- genuinely running across the deploy → the value is in the future → untouched, and its next `TASK_STARTED` re-stamps it from the caller's TTL anyway.

Rows whose `latest_status_at` is itself `NULL` have no basis for a value and are left `NULL`. The TTL is read from the deployment's configuration rather than hardcoded, deliberately: an operator who raised the default because their jobs legitimately run for days would otherwise have every two-day-old **live** task declared expired by a 24h constant. `downgrade` needs nothing extra — dropping the column discards the backfill with it.

**Request parameter** — `POST /builds/{build_id}/tasks/{task_id}/start` takes `claim_ttl_seconds: int | None`. Bounds `[60, 2592000]` (1 minute … 30 days); outside them the request is **422**. Applies to *every* start, not only `claim=true` ones — RUNNING is the claim however it was recorded, and a start that granted no expiry would be exactly the wedge this exists to end. It is echoed into the start event's metadata as `claim_ttl_seconds`.

**Response fields** (all `datetime | None`, all defaulted so they are additive):

| Field | Where |
| --- | --- |
| `latest_status_expires_at` | `FrontierTaskRef` (`roots` / `actionable` / `running` in `GET /builds/{id}/frontier`) |
| `latest_status_expires_at` | `ConcurrencyLimitHolder` (`GET /concurrency-limits/{key}/holders`) |
| `latest_status_expires_at` | the `task_already_running` 409 `detail`, alongside the existing `executor` / `executor_ref` / `latest_status_at` |
| `blocking_status_expires_at` | `FrontierExternalBlocker` (`blocked_by_external` entries) |

Semantics in every case: the instant the holder's claim stops being believable. Meaningful only while the status is RUNNING.

**`null` means "no expiry known" and must be treated as a claim that never lapses — wait, do not fail.** After the backfill that is a much narrower population than it sounds: not "an old claim, so probably dead", but one stamped by a server predating this column and not re-started since (a pre-denormalisation row with no status timestamp to date it, or a straggler from a rolling deploy). Nothing can date those, so releasing them stays an operator action (cancel / retry / evict).

**Setting** — `STARDAG_API_CLAIM_DEFAULT_TTL_SECONDS`, used when the caller supplies none.

## Default TTL: 7 days, and why generous

Client-supplied wins because the caller is the only party that knows how long the execution it is about to spawn can legitimately take; a claim should outlive its execution by a small grace and no more.

For the fallback, the two failure modes are not symmetric. Expiring **late** only delays the self-heal of a task that is already permanently wedged today — the status quo is "never", so any finite value is an improvement. Expiring **early** hands a live task to a second claimant, i.e. a double execution, which is the one thing the claim exists to prevent. So the default is deliberately generous: **seven days**. It was a day in the first revision of this PR — matching `ReaperSettings.idle_for_seconds` — and was raised on review, because a day is *below* the maximum function timeout of some execution backends stardag drives, and a default that can expire a live task is the one failure this must not have. The symmetry with the reaper was the weaker argument: the reaper cancels an idle **build**, which a human can undo, while an expired claim silently authorises a second execution.

Note this default is also what the migration backfills existing RUNNING rows with, so it governs upgrade behaviour as well as runtime.

## Both predicates, not one

The claim check and the concurrency-limit `active` count now use the same rule (`services/claims.py`, expressed once in Python for the FOR-UPDATE-locked row and once in SQL for counting):

```
RUNNING AND (expires_at IS NULL OR expires_at > now())
```

Changing only the first would have preserved the slot leak in the one place nobody reads: the task stops blocking its own re-execution while still consuming the cap it was admitted under. `GET /concurrency-limits/{key}/holders` shares the predicate too, since it exists to explain that count. **Eviction deliberately does not** — an expired claim occupies no slot, but its task still reads RUNNING to every status consumer, and recording the event that ends that is precisely what eviction is for.

## Renewal, stated honestly

Every event that asserts RUNNING (a re-start recording an executor ref, a resume) re-grants the expiry, so there is renewal on lifecycle traffic that already exists, at zero extra cost. **This buys little for the case that matters most:** a long-running task that emits nothing between start and finish gets exactly one expiry, at start. That is not a gap to plug with heartbeats — it is the reason the TTL should be derived from the executor's own timeout rather than being a short lease.

## What this does NOT solve

- **Cross-build blocking does not disappear.** If task U is RUNNING under a dead build and U is not in build X's task set, proving U dead does not let X run it — U is not in X's plan. What the expiry buys is *certainty instead of inference*, nothing more. There is a test pinning exactly this.
- **Executor probing stays.** An expiry is a floor on liveness; the backend's own answer about a call it is running is the truth, and it is better evidence than any timestamp.
- **Build-level reaping stays.** An abandoned *build* is a separate problem from a leaked task claim.
- Nothing in `lib/` changes here, so the stale-build reaper, `stale_running_no_ref_seconds` and the cross-build blocker classification are all untouched.
- A task whose claim lapsed still *reads* RUNNING until some event moves it. The expiry makes it re-claimable and frees its slots; it does not rewrite its status.
- Claims with no `latest_status_at` to backfill from remain unexpiring, and so does anything a pre-column server stamps during a rolling deploy. The cleanup endpoints remain the way to release those.

## Notes for review

- Migration `f6a5873c30fb` (revises `11716b1f4c1a`): schema step autogenerated, backfill hand-written as a deliberate data step. Verified against a real Postgres with `upgrade head` → `downgrade -1` → `upgrade head`, and end to end in the suite — `test_migration_backfills_claims_that_are_already_running` takes a 90-day-old RUNNING claim back through downgrade/upgrade and asserts it comes out with an expiry in the past and is re-claimable, with a companion test that a row without `latest_status_at` stays `NULL` and keeps denying. Not indexed: every reader narrows on `latest_status` first, leaving a comparison over a handful of rows.
- The expiry is derived inside `apply_event_to_task` rather than at the route, so it is a function of the event stream (a replay reproduces it) and no start path can grant an unexpiring claim by forgetting an argument.
- `services/claims.py` refers to the `docs/design/execution-claims-and-liveness.md` note from #218, which is not merged yet.
- 443 tests pass against Postgres, 419 + 25 skipped against SQLite; `pre-commit` (incl. `pyright-stardag-api`) clean.


andhus:
Copilot review triaged — three findings, all accepted, one important.

1. **TTL default is 7 days, PR description says 24h.** The **code is right and the description is stale**. The default was deliberately raised after review; the PR body and CHANGELOG were never updated. This matters because the migration backfill uses the same default, so the description misstates what an upgrade does to existing rows. Description/CHANGELOG to be corrected — not the code.
2. **`live_claim_filter()` docstring says `expires_at > now()` but the code uses `utc_now()`.** Real; application time vs database time is exactly the kind of thing a future reader would change on the strength of a docstring.
3. **Docstring points at a design note not on this branch.** Same cross-branch framing issue as #218 — the note lands in #218.

andhus:
**Now resolved** — all three.

- The docstring said `expires_at > now()`, implying database time; both predicates compare against **application** time, deliberately (the Python half has no session to ask, and the two must share one clock). Now stated, with the reason.
- Dropped the pointer to a design note that is not on this branch.
- **The TTL default was the code being right and the description being stale.** The PR body now says 7 days and why it was raised from 24h on review — a day sits *below* the maximum function timeout of some backends stardag drives, and a default that can expire a live task is the one failure this must not have. It also notes that the same default governs the migration backfill.

PR #221, opened 2026-08-07
refactor(api): denormalise build status so it is filterable, exact and computed once

andhus:
Build status was not a column. `get_build_status` replayed every build-level
event on every read, and `_build_to_response` called it per build. Three
things followed from that, and this PR removes all three.

## What moved

Five denormalised columns on `builds`, holding exactly what the replay
returned:

| column | was |
| --- | --- |
| `latest_status` | `status` |
| `latest_started_at` | `started_at` |
| `latest_completed_at` | `completed_at` |
| `latest_status_triggered_by_user_id` | `status_triggered_by_user_id` |
| `latest_is_resumed` | `is_resumed` |

Plus `ix_builds_environment_status` on `(environment_id, latest_status,
last_active_at)` — same shape and same reasoning as `ix_tasks_environment_status`,
so a single-status page resolves its `ORDER BY` from the index in either
direction.

They are maintained by `apply_event_to_build`, which mirrors
`apply_event_to_task`: applied in the transaction that inserts the event,
on a row locked by `_get_build_for_update`. Every path that creates a
build-level event goes through it — `create_build`, `/complete`, `/fail`,
`/cancel`, `/exit-early`, `/resume`, and `cancel_builds` (bulk cancel and
the reaper). The build is always locked *before* any task rows a cascade
touches, so the lock order is consistent with `_create_task_event`.

One structural difference from the task fold, and it is deliberate: there
is no stickiness. A completed task cannot un-complete, but a build
genuinely goes COMPLETED → RUNNING, because `sd.build(resume_build_id=...)`
on a finished build is supported. So the build fold is plain
last-event-wins.

## The tie-break

The two previous implementations disagreed when a terminal event shared a
`created_at` with a start/resume: the replay ordered on the timestamp and
let the index settle ties; `build_is_running()` read a tie as *not*
running, on purpose, so the reaper would leave it alone.

**The new rule is arrival order** — the order the server committed the
events. That is less a choice than a consequence of folding at write time:
the event insert and the fold share a transaction, so two events are
strictly ordered by their commits even when their timestamps collide.
Equal timestamps mean the proxy lost information the commit order still
has; both old implementations were guessing.

The reaper's conservatism is **consciously changed, and preserved where it
matters**. It can only act on a build whose column says RUNNING *and*
which has since been silent for at least `idle_for_seconds` (floor 60s). A
build whose last committed lifecycle event was a start or resume is
running by every available reading; a build that is running and has gone
quiet for the idle window is precisely what the reaper exists to find. The
case is covered by a test that reproduces the tie and asserts the build is
now selected.

## What was deleted

- `_STATUS_FILTER_SCAN_CAP = 500` and the Python filter path in
  `GET /builds`, along with the `total` that counted matches *within the
  scanned window* while looking like an exact count.
- `build_is_running()` in `services/build_cleanup.py` — a second SQL
  encoding of the same rule — and its call sites, which become
  `Build.latest_status == BuildStatus.RUNNING`. `_TERMINAL_BUILD_EVENTS`
  and `_ACTIVE_BUILD_EVENTS` went with it.
- `get_build_status` itself, rather than leaving a thin wrapper: the whole
  point is that there is one way to get the answer.
- The per-response event scan behind every `BuildResponse`.

## `status` is now exact for every value — but the idle restriction stays

`status` filtering is exact and unbounded for **every** status: true
`COUNT(*)`, server-side pagination, no window. That part of the original
motivation stands.

What did **not** happen is lifting the `status` + `idle_for_seconds`
restriction. An earlier revision of this PR removed the 422 on the grounds
that "only `running` has a SQL predicate" was no longer true. It was
reinstated, because the reason for the restriction turned out not to be the
implementation at all:

An idle filter means **abandoned**, and only a running build can be
abandoned. A finished build satisfies "no activity for N seconds"
trivially and permanently, so a bare staleness query returned every build
that ever completed — sorted stalest-first, which put the oldest completed
builds *above* the running ones an operator is looking for. It also broke
parity with `POST /builds/bulk-cancel`, which only ever cascades over
running builds: the listing is that endpoint's preview, and a preview that
offers rows the action will never touch is worse than none.

So `idle_for_seconds` implies RUNNING, and pairing it with any other status
is a contradiction rather than a narrower query — answered with 422 instead
of the empty result the predicates would produce. Verified against a live
deployment: the query went from 4 builds (2 terminal, sorted top) to 2,
matching the reaper's dry run exactly.

## Migration

`e87efdab31ca`, autogenerated, with the backfill added deliberately: one
pass over build-level events ordered `(created_at, id)`, applying the same
rules as `apply_event_to_build`. The `id` tiebreaker is what makes ties
resolve the same way before and after — `id` is a UUID7, so ordering by it
is ordering by arrival. The two NOT NULL columns take a `server_default`
so the ALTER succeeds on existing rows, dropped again afterwards.

Verified against Postgres 16: `upgrade head` → `downgrade -1` →
`upgrade head`, then `alembic check` clean. Eleven seeded event streams —
every terminal, resume-then-terminal, start-after-resume, both tie
orderings, and a build with no events at all — compared field by field
against the old replay; all matched.

## Tests

`app/stardag-api/tests/test_build_status_denormalisation.py` (new) covers
`apply_event_to_build` including the tie-break, every lifecycle endpoint
asserted **against the `builds` row** rather than the response body (a
handler that forgot the fold but recomputed the response some other way
would pass an endpoint test and leave the row wrong), a full-lifecycle
row-versus-replay check, exact `total` and server-side pagination for
non-`running` statuses, and the reaper selecting from the column.
`test_status_denormalisation_concurrency.py` gains the build mirror of the
task race tests, including a deterministic demonstration of the torn row
the lock prevents (COMPLETED, with a `completed_at`, yet flagged resumed)
and a 40-way stress test through the real handlers.

The two tests that asserted the approximate `total` and the 422 are
rewritten to assert the exact behaviour that replaced them.

`478 passed, 1 skipped` (SQLite + Postgres); pyright and the pre-commit
hooks clean.

## Net lines

`src/`: +347 / -342 — net **+5**, and much of the addition is the comment
explaining the tie-break and the lock. `tests/`: +842 / -41. Migration and
CHANGELOG: +234 / -6.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -91,12 +91,9 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
   predicate, not a second implementation) — so a client can list what the
   reaper would cancel before cancelling it. Because it is a real SQL
   predicate, `total` is an exact `COUNT(*)` and pagination is server-side
-  and unbounded, unlike the derived-`status` filter's bounded window.
-  Ordering flips to stalest-first when it is given. `status=running` is
-  accepted alongside it and then also resolves in SQL, so the
-  abandoned-build query stays exact past the 500-candidate scan cap; any
-  other `status` value combined with `idle_for_seconds` is rejected with
-  422 rather than served an approximate total that looks exact.
+  and unbounded. Ordering flips to stalest-first when it is given. It
+  combines with any `status` value — see the denormalisation entry below,
+  which made that true for the non-`running` ones too.
 - **Optional unattended sweep** (`STARDAG_API_REAPER_ENABLED`, off by
   default) runs the same operation on a timer inside the API process. Note
   that every replica runs its own timer with no leader election; cancellation
@@ -153,6 +150,42 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
     since — and those still need an operator (cancel, retry, evict) to
     release. Executor probing is unaffected and remains the better evidence
     where it is available.
+- **Build status is now a column, so `GET /builds?status=` is exact for
+  every status** ([#208](https://github.com/stardag-dev/stardag/issues/208)).
+  It was derived by replaying the build's build-level events on every read,
+  which had three consequences, all now gone: every `BuildResponse` cost an
+  event scan; the `status` filter could not run in SQL, so it scanned the 500
+  most-recently-active candidates, filtered in Python, and returned a `total`
+  that was **the matches within that window** while looking like an exact
+  count; and anything needing an unbounded "is this build RUNNING?" — the
+  stale-build reaper — had to carry a second SQL encoding of the same rule,
+  which disagreed with the replay when a terminal event shared a timestamp
+  with a start/resume.
+
+  Five denormalised columns on `builds` (`latest_status`,
+  `latest_started_at`, `latest_completed_at`,
+  `latest_status_triggered_by_user_id`, `latest_is_resumed`) now hold exactly
+  what the replay returned, folded in-transaction by every build lifecycle
+  path — the same pattern already used for `tasks.latest_*`. Backed by a new
+  `(environment_id, latest_status, last_active_at)` index. Migration
+  backfills existing builds by replaying their events, so no build changes
+  status across the upgrade.
+
+  - `status` filtering is a plain column predicate: exact `COUNT(*)`,
+    server-side pagination, no window and no cap, for every status.
+  - `status` combined with `idle_for_seconds` **no longer 422s** for
+    non-`running` statuses. That restriction existed only because `running`
+    was the sole value with a SQL predicate; "failed, and idle for a week" is
+    now a real query.
+  - **Tie-break:** two build-level events sharing a `created_at` resolve in
+    _arrival_ order — the order the server committed them — because the fold
+    shares a transaction with the event insert. Equal timestamps mean the
+    timestamp lost information the commit order still has, and both previous
+    implementations were guessing (the replay arbitrarily, the reaper by
+    always reading a tie as "not running"). The reaper is unaffected in
+    practice: it still only touches builds that are RUNNING _and_ have been
+    silent for at least `idle_for_seconds`.
+  - No response field changed shape or meaning.
 
 ### SDK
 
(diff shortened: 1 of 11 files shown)


andhus:
Copilot review triaged — five findings. Two describe a **deliberate reversal** that landed after this review's premise, so the resolution there is to fix the PR description rather than the code.

**On `idle_for_seconds` forcing RUNNING (two comments).** The review is right that the code contradicts this PR's description and CHANGELOG — but the code is now the intended behaviour and the prose is stale. An idle filter means *abandoned*, and only a running build can be abandoned: a finished build satisfies "no activity for N seconds" trivially and permanently, so a bare staleness query was returning every build that ever completed, sorted stalest-first, i.e. the oldest completed builds ranked above the running ones an operator is looking for. It also broke parity with `POST /builds/bulk-cancel`, which only ever cascades over running builds — a preview that offers rows the action will not touch. Verified against a live deployment: the query went from 4 builds (2 of them terminal, sorted top) to 2, matching the reaper's dry run exactly. **The 422 guard stays; the description and CHANGELOG need updating.**

**Migration backfill is O(#builds) in memory.** Accepted and the most substantive of the five — this runs on real installations at upgrade time and should stream or batch rather than building the full `states` dict.

**Test seeds `BUILD_FAILED` without folding it into `Build.latest_*`.** Accepted: that state is impossible in production after this refactor, so the fixture can mask exactly the event-stream-vs-column bugs the refactor is meant to eliminate.

**Stray leading quote in a docstring.** Accepted.

andhus:
**Now resolved** — and two of the five were the description being stale, not the code.

- **Migration backfill now streams.** Ordered by `build_id` first (the replay only depends on ordering *within* a build), state flushed on id change, batched updates. One trap worth knowing: setting `stream_results` on the *connection* leaves Alembic running the rest of the migration on a streaming connection, where rowcount comes back -1 and Alembic aborts with "expected to match one row". It has to be statement-scoped. **Only the Postgres run caught this** — the SQLite suite passed either way.
- Concurrency fixture no longer seeds a build-level terminal event without folding it into `latest_*`; it uses a task-level event, so the stream and the columns agree.
- Docstring rephrased to avoid the stray leading quote.
- **The 422 stays.** The PR body has been rewritten: an earlier revision did remove it, and that was reverted because the reason for the restriction was never the implementation. Idle means *abandoned*; only a running build can be. Details in the updated description.

PR #222, opened 2026-08-07
refactor(reactive): decide blocker liveness from claim expiry instead of inferring it

andhus:
SDK side of the claim-expiry work: the server now gives every execution
claim an expiry (`latest_status_expires_at`, `NULL` = "never lapses"), and
an expired claim is re-claimable and no longer counted against concurrency
limits. That replaces the liveness *inference* a tick was doing with a
fact it can read.

## The final decision table

`_classify_external_blockers`, for a blocker of a build that has nothing
actionable and nothing running:

| in this build's task set | blocking status | evidence consulted | action |
| --- | --- | --- | --- |
| yes | any | none | **ignored** for wait/fail (already in `actionable`/`running`/`status_counts`); still named in the message |
| no | RUNNING | claim expiry, in the future | **wait** — someone is executing it |
| no | RUNNING | claim expiry, in the past | **fail** — the claim is provably abandoned |
| no | RUNNING | no expiry (`NULL`) | **wait**, unbounded, logged as such |
| no | not RUNNING | owning build still live (`running`/`pending`) | **wait** — it is going to schedule it |
| no | not RUNNING | owning build terminal | **fail** |
| no | not RUNNING | no owning build recorded | **fail**, without a lookup |
| no | not RUNNING | owner status unresolvable / unreported | **fail** — unknown is not evidence of life |

The RUNNING half is where the collapse happens: three rows and **zero**
requests, down from a table that also weighed a staleness bound. The
owning build is not consulted for a RUNNING blocker in *either* direction —
a live build proves nothing about one of its claims, and a terminal one
does not release them.

`NULL` is a deliberate row, not an omission. It is the server's own
encoding of "never lapses" (older server, or a start predating the
column), and reading missing evidence as death would fail builds whose
blocker is perfectly alive — the exact spurious failure #208 A1 removed.
The window is self-closing as RUNNING tasks turn over, and the tick logs
that the wait cannot be shown to end, naming the cancel that breaks it.

## What this deliberately does not solve

**Expiry speaks only about RUNNING.** A SUSPENDED or PENDING task holds no
claim, so the server clears the expiry with it and there is nothing to
read. Abandoned-SUSPENDED is a real wedge, so the owning-build lookup —
and its per-pass memo — **survives, for the non-RUNNING half only**. Not
kept out of caution: it is the only evidence that exists there. What a
live owner no longer buys is a deadline; a build that goes silent without
transitioning is reaped server-side, not guessed at here from a task's
age. Stated in the code and covered by a test that asserts the lookup
still happens for a SUSPENDED blocker.

**A dead blocker does not unblock the build.** If the blocking task is not
in this build's task set, proving it dead does not let this build run it —
the build still fails. What changes is certainty and the message, which
now says the claim lapsed and that the blocker has to be released under
the build that owns it. Nothing here claims expiry makes cross-build
blocking disappear.

**Executor probing is untouched.** An expiry is a floor on liveness; the
backend's own answer is the truth. `_resolve_running` keeps its
target-first, then-probe precedence exactly as it was wherever a ref
exists.

## The TTL derivation

`claim_ttl_seconds(task, executor)` = the executor's own wall-clock limit
+ a 15-minute grace, clamped to the server's `[60, 2592000]`; `None` (so
the registry's default) when the backend enforces no limit.

Granting an expiry on *every* start maximises healing, but it also makes a
task that outlives its TTL stealable **while alive** — a duplicate
execution, i.e. the thing claims exist to prevent. Deriving the TTL from
the limit the backend itself enforces is what keeps that from being a real
risk: the backend will have killed the execution before its claim lapses,
so the only claims that can lapse belong to executions already dead. A
generic server-side default cannot make that promise, which is why "the
server has a default" is not a reason to omit it.

New optional hook `TaskExecutorABC.execution_timeout_seconds` (default
`None`). `ModalTaskExecutor` answers it from the deployed
`FunctionSettings(timeout=...)` of the worker a task routes to, passed
into the tick's executor at `finalize()` — the only place those settings
exist. The grace covers the three ways the two clocks differ: the claim is
recorded at *acquire* time, before the spawn, so it absorbs queueing and
cold-start latency; a backend does not kill at the instant of timeout; and
clocks are not identical.

The TTL is sent on **every** start, not just the claiming one — the
post-spawn ref-recording start and the worker's own self-reported start
included. A worker's TASK_STARTED is a start like any other and would
otherwise re-stamp the claim with the server default, undoing the
derivation; forwarding it (on the existing env-override channel) also
re-bases the expiry on the moment execution actually begins.

## Deleted

- `TickConfig.stale_external_blocker_seconds` and the bounded wait built on it.
- `TickConfig.stale_running_no_ref_seconds` **and**
  `ClaimConfig.stale_running_no_ref_seconds` (different defaults, same guess).
- `_blocker_is_stale`, the two staleness rows in the classification, and the
  `_claim_holder_is_stale` twin in `build/_concurrent.py`.
- `StartClaimResult.latest_status_at`, which existed for exactly one caller:
  the bound above. Replaced by `latest_status_expires_at`.

**The RUNNING-without-ref window was checked, not assumed away.** It is
still reachable — the claiming start is recorded before the spawn, so a
tick that dies in between leaves exactly that shape — and still needs
handling, because such a task holds its concurrency-limit slots forever.
So the minimum is kept: fail it when its claim has **lapsed**, leave it
otherwise (the spawn-in-progress window of a perfectly healthy tick is
indistinguishable from here). What is gone is the locally configured guess
at how long is too long.

## Net line count

`+986 / -352` across 21 files (net **+634**), which is not the net
deletion this was scoped as. Honest breakdown:

- The classification collapse and the knob removals are net negative
  (`TickConfig` alone loses ~40 lines of field + rationale, plus
  `_blocker_is_stale`, the `ClaimConfig` field, the two staleness rows and
  their tests).
- The TTL derivation is the additive half and is new behaviour, not
  plumbing: `claim_ttl_seconds` + its rationale, the executor hook, the
  Modal wiring and worker forwarding (`+79` in `modal/_app.py` alone), and
  their tests (`+80`).
- Docs and CHANGELOG account for `+123`.

`src/` alone is about `+450 / -200`. Sizing this as a deletion assumed the
TTL half was plumbing; it is not.

## Verification

From `lib/stardag`:

```
uv run pytest tests -q -m "not modal_live"   →  1174 passed, 29 deselected
uv run pyright src tests                     →  0 errors, 0 warnings
```

(1160 on the base; +14 net new tests.) `uvx pre-commit run --files` over
every changed file passes. No `modal_live` test was run.

Based on #219. Against an older server every path here degrades to today's
behaviour: no expiry is reported, so nothing is ever judged lapsed, and
`claim_ttl_seconds` is an unknown query param the server ignores.


andhus:
Copilot review addressed — both findings were real, and the first is a genuine correctness bug on a supported interpreter.

**`fromisoformat` and `Z`.** Confirmed: `requires-python = ">=3.10"`, and `fromisoformat` only learned to accept a trailing `Z` in 3.11. The registry serialises claim expiries in exactly that form (`ClaimDenial.latest_status_expires_at` is a `str` straight off the payload — distinct from `FrontierTaskRef`, which is a parsed `datetime`). On 3.10 every expiry failed to parse, and an unparsed expiry reads as *not lapsed* — the dangerous direction, leaving the loser waiting out a claim nobody holds. Normalised before parsing.

I tried to add a discriminating regression test and could not: this repo runs 3.12 locally, where `Z` parses natively, so the test passes with or without the fix. Only the 3.10 leg of CI exercises the normalisation. Rather than ship a test that looks like coverage and is not, I left it out — flagging that explicitly.

**`STARDAG_CLAIM_TTL_SECONDS` range.** Real. The parse was already tolerant of garbage for a good reason, but a syntactically valid `0` sailed through to `task_start` and came back 422, losing the whole lifecycle report to exactly the failure the tolerance exists to prevent. Non-positive values are now dropped with a warning.

PR #223, opened 2026-08-07
feat(api): expose per-build task attempt counts so schedulers can retry

andhus:
Server half of a task-level retry policy for reactive scheduling
([#208](https://github.com/stardag-dev/stardag/issues/208)). A follow-up
SDK PR consumes it; nothing here changes existing behaviour.

Based on `refactor/denormalise-build-status` (#221).

## Why

In reactive mode the scheduler records a failed execution and never
respawns it — *"retries are the execution backend's job"*. But a backend's
function-level retries only cover exceptions raised **inside** a container
that started. They do not cover:

- a spawn that failed before the container existed,
- an OOM kill or a preemption,
- a worker that died after writing partial output.

There is no `max_attempts` anywhere in the SDK, so under `FAIL_FAST` a
single transient failure kills the whole build. Closing that needs a
**durable** attempt count: a tick is short-lived and cannot remember how
many times it has tried.

## The field

**`attempt_count: int`** — how many times execution has been started for
this task **since its build's most recent `BUILD_RESUMED` event**, or
since the build began if it has never been resumed. `0` means "not
attempted in this round".

**Presence is guaranteed.** Every response of this API that declares the
field always populates it — it is a non-optional `int` on the schema, and
every route that returns one of these models computes it (including the
concurrency-limit eviction route, which spends an extra query purely so
there is no "was it computed here?" case). A consumer that sees the key
missing is therefore talking to a server older than this PR, and *that* is
the only reason to model it as optional client-side. Do not default a
missing key to `0`: absence means "unknown, don't enforce", and collapsing
it into `0` would make an unbounded retry loop unavoidable.

Exposed on:

| surface | model | notes |
| --- | --- | --- |
| `GET /builds/{id}/frontier` | `FrontierTaskRef` | on all of `actionable`, `running`, `roots` |
| `GET /builds/{id}/tasks` | `TaskWithStatusResponse` | for UI/CLI; this model has no other consumer, and that consumer is per-build |
| every task lifecycle event | `TaskEventResponse` | `start` / `fail` / `retry` / `complete` / … , **including the event just recorded** |

The event responses are there so a caller that has just recorded a failure,
or won a claiming start, can apply its budget without a second round-trip —
and against a count that is authoritative rather than a snapshot some
racing scheduler may have moved on from.

**Not** added to the environment-global task responses (`TaskResponse`,
`GET /tasks`), where "how many attempts" would have no build to be about.

## The double-start rule

`attempt_count` counts **attempts, not `TASK_STARTED` events**. Engines
routinely emit several starts per execution:

- **reactive** (`_act_on_frontier`): an *acquiring* start — the atomic
  claim / limit-slot acquisition, before the spawn, with no executor ref —
  and then a second start carrying the ref once the spawn returned.
- **resident** (`_concurrent`): the same claim-then-ref pair, plus the
  worker's own self-reported start when the executor reports lifecycle,
  which lands minutes later under cold start. Three starts, one execution.
- **sequential / unclaimed**: exactly one start.

Counting events would make the same task look like it had 1, 2 or 3
attempts depending on which engine and which executor ran it, and would
exhaust any retry budget on the first try.

The rule instead counts **transitions into RUNNING**: a start begins a new
attempt unless the previous *status-affecting* event was itself a start.
A run of consecutive starts collapses to the one execution it describes,
whatever the engine's reason for re-recording. This is engine-agnostic by
construction — it needs no knowledge of who emits how many starts, only
that nothing else happened to the task in between.

Status-neutral events (`TASK_REFERENCED`, `TASK_PENDING`,
`TASK_WAITING_FOR_LOCK`) are excluded from "previous", so one landing
inside an attempt's acquire→spawn window cannot split it in two. They
cannot change status, so they cannot end an attempt either.

Pinned by `tests/test_task_attempts.py`, which asserts `1` for all three
shapes above, plus the neutral-event case, plus that a start *after* a
terminal event does begin attempt 2.

## The round window: a resume resets, a retry does not

This is the part that makes the budget usable, and the two halves are
easy to conflate.

**`BUILD_RESUMED` starts a new round, and the count restarts at zero.**
`build_trigger(..., build_id=<existing>, reactive=True)` is fully
supported and is the recommended way to pick a failed reactive build back
up. It does **not** mint a new build: `_trigger_reactive` resumes *this*
one and its discovery calls `task_retry` on the failed tasks in it. If the
count spanned the build's whole history, the budget would already be spent
the moment the user asked for another go, and resuming would additionally
require raising `max_attempts` — for the one action anybody reaches for
after a failure.

`BUILD_RESUMED` is the right cut rather than a special case:

- it is exactly the durable marker of "another round was asked for";
- it needs no new event type, column or marker — the event already
  exists and is already emitted on that path;
- it keeps the count per-build, so nothing about the scoping argument
  below changes;
- the server only emits it when the build has activity beyond
  `BUILD_STARTED`, so a **first** trigger is unaffected (verified: a
  no-op resume before any task activity records nothing and creates no
  round boundary);
- repeated re-triggers do hand out fresh budgets, but that is a user
  deciding to try again — visible and user-driven, not a scheduler
  looping.

**`TASK_RETRIED` on its own does not reset.** A scheduler enforcing a
budget retries a failed task *through* that endpoint — a `FAILED` task is
not in the frontier's actionable set until something flips it back to
`PENDING` — so a counter cleared by it would be cleared by every
enforcement of the budget it defines, and `max_attempts` would never bind.

The distinction is what lets a scheduler tell **"this round is out of
attempts"** (a real message to show a user who clicked Retry) from **"the
user asked for another round"** (a real reset).

One implementation detail worth naming, because getting it backwards is
silent: the round cutoff is applied **before** the `LAG`, not after. If it
were applied afterwards, the last start of the previous round would still
be the first post-resume start's predecessor, collapse it, and a resumed
build would report zero attempts however many times it tried.

## Per-build scoping

Unlike the denormalised `latest_*` fields it travels with — environment-
global, because a completed task is completed for everyone — this is
scoped to **one build** (and now one round of it). A task that spent two
attempts in an earlier build must arrive in a fresh trigger with a full
budget, or a lifetime budget would make it permanently unschedulable in
every future build in the environment.

Because it differs from everything around it, it is stated explicitly in
the `FrontierTaskRef` field comment, in the frontier endpoint docstring,
and on the deriving function.

## Derive vs denormalise

**Derived**, and there was not much of a contest. Attempts are per-build
and per-round, and there is no per-`(build, task)` row in the schema to
hang a column on. Denormalising would mean a new table, a fold on every
start path (three of them, all racy), a backfill, and — with the round
window — a reset fanned out across every task of a build on each resume.
All to replace one grouped query over an index that already exists
(`ix_events_build_task_type`). Deriving also keeps the rule in exactly one
place, expressed declaratively, where it cannot drift from the event log.

**No migration**, no new column, no backfill.

## Query shape

One grouped statement: a `LAG` window function partitioned by task and
ordered by `(created_at, id)`, filtered to the status-affecting event
types and to the round, with a null-safe inequality so the first event in
a stream still counts. The round cutoff is a correlated scalar subquery
(`max(created_at)` of the build's `BUILD_RESUMED` events, a point lookup
on `ix_events_build_task_type` since build-level events carry
`task_id IS NULL`), so windowing costs no extra round trip.

`LAG`, `IS DISTINCT FROM` and the correlated subquery all render
differently per dialect, so there is a Postgres pass over all three rules
(verified against a throwaway 16 instance).

## Cost

**Frontier: +1 query, from 5 to 6** — measured, not estimated, and
unchanged by the round window. It is a single grouped query, never one per
task: the frontier is re-read on every linger poll (~3 s per active
build), so an N+1 there would be the worst possible place for one. The
scan is bounded to the union of task pks the response actually carries
(`actionable ∪ running ∪ roots`), so the added cost tracks the size of the
response rather than the size of the build's event history — and a
frontier with no tasks to report issues **no** extra query at all.

`GET /builds/{id}/tasks`: +1 grouped query, mirroring the `artifact_count`
aggregate beside it.

Task event responses: **+1** small indexed lookup for the round cutoff.
The tally itself still rides on the per-build replay
`get_task_status_in_build` already performs, but the cutoff cannot be
pushed into that query — the *status* fold has to see the whole stream (a
task completed before a resume is still completed), so only the tally is
windowed. That replay also gained an explicit `id` tiebreak on its
ordering so equal timestamps replay deterministically, which the attempt
rule is sensitive to.

The concurrency-limit eviction route spends one extra query so the field is
never null anywhere (see **Presence is guaranteed** above). That route is
admin recovery and runs at human frequency.

## Verification

```
468 passed, 29 skipped        # SQLite (default)
495 passed,  1 skipped, 1 failed   # with STARDAG_API_TEST_DATABASE_URL → Postgres 16
```

The one Postgres failure is
`test_claim_expiry.py::test_migration_backfills_claims_that_are_already_running`,
which is **pre-existing on the base branch** and unrelated to this PR —
confirmed by reverting `src/` to the base commit and reproducing it
identically. This PR touches neither migrations nor claim expiry.

`pre-commit` clean on every changed file, including `pyright-stardag-api`.

Purely additive: every existing field keeps its exact semantics.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,10 +20,12 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
   tick is short-lived and cannot remember what it already tried.
 
   `attempt_count` is how many times execution has been started for a task
-  **in that build**, exposed on `FrontierTaskRef` (all of `actionable`,
-  `running` and `roots`), on `GET /builds/{id}/tasks`, and on every task
-  lifecycle event response so a caller that has just recorded a failure or
-  won a claiming start can apply its budget without a second round-trip.
+  **since its build's most recent `BUILD_RESUMED` event** — since the
+  build began, if it never was. Exposed on `FrontierTaskRef` (all of
+  `actionable`, `running` and `roots`), on `GET /builds/{id}/tasks`, and on
+  every task lifecycle event response so a caller that has just recorded a
+  failure or won a claiming start can apply its budget without a second
+  round-trip. Always populated wherever it is declared.
 
   Three things it is easy to assume and get wrong:
 
@@ -37,17 +39,28 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
   - **The scope is the build, not the environment** — unlike the
     `latest_*` fields it travels with. A task that spent two attempts in
     an earlier build arrives in a fresh trigger with a full budget.
-  - **A retry does not reset it.** A scheduler retries a failed task
-    _through_ `POST .../retry`, so a resetting counter would be cleared by
-    every enforcement of the budget it defines. Within one build the
-    number is a lifetime budget; a re-trigger gets a new build.
+  - **A resume resets it; a retry does not.** `build_trigger(...,
+build_id=<existing>, reactive=True)` is the recommended way to pick a
+    failed reactive build back up, and it does _not_ mint a new build: it
+    resumes this one and retries the failed tasks in it. Counting over a
+    build's whole history would therefore leave the budget spent the
+    moment the user asked for another go, and resuming would mean also
+    raising `max_attempts`. `BUILD_RESUMED` is the durable marker of
+    "another round was asked for" — and the server already skips it for a
+    build with no activity beyond `BUILD_STARTED`, so a first trigger is
+    unaffected. `TASK_RETRIED` on its own does not reset, because a
+    scheduler retries _through_ that endpoint: a counter cleared by it
+    would be cleared by every enforcement of the budget it defines. So a
+    bare retry against a spent budget is a real "this round is out of
+    attempts", and resuming is the answer to it.
 
   Derived from the event log rather than denormalised — attempts are
-  per-build and there is no per-(build, task) row to hang a column on — as
-  one grouped query bounded to the tasks being reported. No migration; the
-  frontier costs exactly one extra query, and none at all when it has no
-  tasks to report. Purely additive: every existing field keeps its exact
-  semantics.
+  per-build, per-round, and there is no per-(build, task) row to hang a
+  column on — as one grouped query bounded to the tasks being reported,
+  with the round cutoff riding along as a correlated subquery. No
+  migration; the frontier costs exactly one extra query, and none at all
+  when it has no tasks to report. Purely additive: every existing field
+  keeps its exact semantics.
 
 - **Reactive scheduler tick summaries are now persisted per build**
   (`POST`/`GET /builds/{build_id}/tick-summaries`). A reactive build is
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,47 @@ For detailed SDK migration guides, see [RELEASE_NOTES.md](RELEASE_NOTES.md).
 
 ### Registry API
 
+- **Task refs now carry `attempt_count`, so a scheduler can run a retry
+  policy** ([#208](https://github.com/stardag-dev/stardag/issues/208)).
+  Reactive scheduling records a failed execution and never respawns it —
+  retries are the execution backend's job — but a backend's function-level
+  retries only cover exceptions raised _inside_ a container that started.
+  A spawn that failed before the container existed, an OOM kill, a
+  preemption, or a worker that died after writing partial output are
+  invisible to them, so under fail-fast a single transient failure killed
+  the whole build. Deciding otherwise needs a _durable_ attempt count: a
+  tick is short-lived and cannot remember what it already tried.
+
+  `attempt_count` is how many times execution has been started for a task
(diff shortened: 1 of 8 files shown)


andhus:
Copilot review triaged — three findings, all accepted.

1. **`last_build_resumed_at` does not constrain `task_id IS NULL`.** Real, and the comment above it claims an index point-lookup the query does not actually express.
2. **CHANGELOG indentation.** Real — line falls out of the nested list.
3. **"+1, unconditionally" comment.** Real: `get_attempt_counts_in_build` returns early without touching the DB when there are no tasks, and that comment exists specifically to justify the extra query cost.

andhus:
**Now resolved** — all three.

`last_build_resumed_at` now constrains `task_id IS NULL`, so it is the index seek its docstring claimed rather than a scan — it runs as a correlated subquery on the frontier, the hottest read in the system. Plus the query-count comment (+1 or 0, since the attempt query is skipped when there are no tasks) and the CHANGELOG indent.

PR #224, opened 2026-08-07
perf(reactive): bounded concurrent fan-out and discovery in the scheduler tick

andhus:
Two places in the reactive scheduler did sequentially what the resident
engine already does concurrently. Both are removed here. This is a
throughput change only — behaviour is identical for any DAG that
completes today.

## Before / after: one wide layer

Per actionable task, `_act_on_frontier` does a task-store read, an
acquiring `task_start_claim`, an executor spawn, and a `task_start`
recording the executor ref. That is 2–3 registry round-trips plus one
spawn.

| | before | after |
| --- | --- | --- |
| 5,000-task layer | ~12,000 HTTP calls, one after another, in one container | same calls, ≤ 50 in flight |
| in-flight bound | 1 | `TickConfig.max_concurrent_actions` (default 50) |
| per-pass spawn cap | none — "however many are actionable" | `TickConfig.max_spawns_per_tick`, derived by default |

Discovery had the same shape: `discover_and_register_aio` walked the DAG
with a recursive `await task.complete_aio()` and no concurrency, while
`build/_concurrent.py` has always done the same walk 50 checks at a time.
That gap was not a once-per-build cost — this walk runs at every reactive
trigger **and** in every worker registering dynamically yielded
dependencies, so it was paid on the hot path of every dynamic-dependency
yield.

## The bounds and where the defaults come from

- **`TickConfig.max_concurrent_actions` — default 50.** Not a new number:
  it is the resident engine's own `max_concurrent_discover` default,
  which has been driving the same registry from a single process since
  long before ticks existed. Keeping one number is also what stops the
  two paths drifting apart again.
- **`discover_and_register_aio(max_concurrent_discover=...)` — default
  50**, for the same reason, so the two discovery paths stop being 50×
  apart.
- **`TickConfig.max_spawns_per_tick` — derived, not picked.** The number
  that matters is a duration, not a count. One spawn costs a bounded
  amount of round-trips (`_SECONDS_PER_SPAWN`), `max_concurrent_actions`
  of them run at once, and the batch has to fit inside a fraction of the
  container's wall clock, so the cap is
  `fraction × timeout × concurrency / seconds_per_spawn`, clamped.

  *Which* timeout is resolved down a ladder, most specific first:

  | rung | input | note |
  | --- | --- | --- |
  | 1 | `max_spawns_per_tick` | explicit override, always wins |
  | 2 | `TickConfig.tick_timeout_seconds` | **this container's** wall clock — the quantity the cap is actually about |
  | 3 | executor's `execution_timeout_seconds` (min across candidates) | fallback *proxy*, labelled as one |
  | 4 | `_DEFAULT_MAX_SPAWNS_PER_TICK` | nothing is known anywhere |

  Rung 2 is filled in automatically in a Modal deployment:
  `_tick_function_timeout_seconds` reads the `timeout` the `tick` function
  is actually registered with, applying the same
  `tick_settings or builder_settings` fallback `finalize` uses — reading
  `tick_settings` alone would report "unknown" for every app that does not
  configure the tick separately, which is the common case.

  Rung 3 exists only for callers that cannot know their own limit. It
  measures how long the spawned *executions* may run, a different quantity
  that can differ by orders of magnitude in the direction that hurts (a
  24-hour worker under a 5-minute tick), so the tick's log line says when
  it is on that rung. `TaskExecutorABC.execution_timeout_seconds` is
  otherwise unchanged — it remains exactly right for the claim TTL, which
  is what it was added for.

  `tick_timeout_seconds` is deliberately **not** in
  `_TICK_KWARGS_ALLOWED`: a deploy-time fact about the container has no
  business being frozen into a build's persisted tick config, where it
  would go stale on the next redeploy. It is applied as a *default* rather
  than an override so the one caller with a legitimate reason to narrow it
  can — see the watchdog note below. `max_spawns_per_tick` and
  `max_concurrent_actions` *are* accepted as `tick_kwargs`.

Bounded everywhere: unbounded fan-out at the registry would only move the
failure from the tick's clock to the connection pool.

## Ordering guarantees preserved, and how they are tested

Concurrency reorders tasks against each other. It does not reorder
anything *within* a task, and it does not touch the claim.

- **Acquiring start → spawn → ref-recording start** stays a sequence
  inside one coroutine, so a denied task never reaches `submit_detached`
  and never occupies a worker, and no executor ref is ever recorded for
  an execution that does not exist yet.
  `test_ordering_holds_per_task_under_concurrency` asserts
  `claim < spawn < ref-start` for all 40 tasks of a layer, having first
  asserted the layer really did interleave.
  `test_denied_task_never_reaches_a_worker` runs a 10-wide layer against
  a one-slot limit: exactly one spawn, and no denied task appears in the
  spawn log.
- **Claims** are unaffected: each task is claimed at most once per pass
  and the tick holds the build's scheduler lease, so these coroutines are
  not racing each other — the claim still arbitrates against *other*
  builds, exactly as before. Spelled out in a comment, because it is the
  first thing the next reader will worry about.
- **Counters** are mutated with a bare `+=` between awaits, which is
  atomic on a single event loop.
  `test_counters_stay_accurate_under_concurrency` drives spawned,
  self-healed and two flavours of recorded failure through one concurrent
  pass and checks each is counted exactly once.
- **Discovery's post-order** survives because the concurrent phase
  records only facts (is this complete, what are its deps) and a second,
  sequential, I/O-free phase replays the original recursion over them.
  `test_matches_the_serial_walk_exactly_for_a_diamond` compares against
  the pre-change walk, kept verbatim in the test file as the reference
  implementation, element for element:
  `incomplete`, `previously_completed`, `retried` and the registration
  order. `test_post_order_holds_under_concurrency` and
  `test_shared_dependency_is_checked_and_registered_once` cover the
  diamond and the single-check/single-register invariant. Bulk
  registration stays sequential across chunks — chunk *n* may hold chunk
  *n+1*'s dependencies.

**The bounds themselves are pinned**, which is the point of the PR:
`test_wide_layer_spawns_concurrently_within_the_bound` and
`test_completion_checks_run_concurrently_within_the_bound` assert the
observed peak in-flight equals the configured bound. Remove the
semaphore and the peak becomes the whole layer; remove the TaskGroup and
it becomes 1. Either way the test fails loudly.

## Spawn-cap behaviour

Truncation is logged, never silent, and is not a stall. A pass that
truncated necessarily *acted* on everything it did attempt, so the
existing "acted → re-evaluate on a fresh frontier" path takes the next
batch immediately — no wake-up, no linger, no watchdog. The one case
where it waits instead is the one where waiting is correct: every
attempted task was denied a concurrency-limit slot, so what the build
needs is a slot to free up, and re-acting at once would be a hot loop
against the registry.

`test_cap_truncates_and_the_tick_re_acts_immediately` makes that
observable by setting `linger_seconds=0` — the linger loop returns on its
first check, so 30 tasks under a cap of 10 can only all get spawned if
the tick genuinely re-acted, three passes in a row.
`test_ticks_timeout_bounds_a_real_pass` does the same end to end from a
`tick_timeout_seconds`, proving the plumbing reaches the fan-out rather
than only being readable in `_spawn_cap`.

Every tick logs its cap **and which rung produced it**, once per tick
(`test_the_cap_and_its_source_are_logged_once_per_tick`). Three of the
four rungs yield plausible-looking numbers from very different inputs, so
a truncating tick is only diagnosable if the log names the input.

**The watchdog needed one thing.** It sweeps every running build
*sequentially inside one container* (`_run_watchdog_sweep` calls the tick
in-process, it does not spawn one container per build), so every build
would otherwise size its fan-out as though it owned the whole timeout and
the first wide build would spend the lot. Each build is now handed a
proportional share of the container's budget alongside the existing
`linger_seconds: 0`. Truncating there is cheap and self-correcting — the
watchdog is a safety net, builds are normally driven by their own ticks,
and a truncated pass re-acts on a fresh frontier immediately.

## Verification

```
uv run pytest tests -q -m "not modal_live"   → 1202 passed, 29 deselected
uv run pyright src tests                     → 0 errors
uvx pre-commit run --files <every changed file>  → all hooks pass
```

1174 → 1202: 28 new tests. The only pre-existing tests touched are the
watchdog-sweep and tick-config assertions, extended for the new budget
plumbing.


andhus:
Copilot review triaged — one finding is a false positive, one is real and is queued.

**`_TICK_KWARGS_ALLOWED` defined twice.** Not correct — it is assigned once (line 372). The other three occurrences the review points at are references inside a docstring and the validation code, not assignments. No shadowing, and the new fan-out kwargs are in the allowlist. Verified by grep on the branch; no change made.

**`_tick_function_timeout_seconds` truthiness on `0`.** Real — a configured `0` is indistinguishable from unset. Not yet fixed; tracked with the rest of the outstanding review items.

This branch has been rebased onto the updated stack below it and its suite is green (1252 passed).

andhus:
**Now resolved** — the real one.

`_tick_function_timeout_seconds` tested the timeout for truthiness, so a configured `timeout=0` read as "not declared" and sent the spawn cap to a different fallback rung than the one the function was registered with. Now `is not None`, with a test.

I initially also switched the *settings-level* fallback to `is not None` and the suite caught it: an empty `tick_settings` dict genuinely does declare nothing, and `finalize` resolves it the same way. Reverted to truthiness there.

The duplicate-definition finding remains a false positive — `_TICK_KWARGS_ALLOWED` is assigned once.

PR #225, opened 2026-08-07
feat(reactive): task-level retry policy for failures the backend cannot retry

andhus:
Reactive builds had no task-level retry policy. `_act_on_frontier` recorded a
failed execution and never respawned it — *"retries are the execution
backend's job"*. They partly are: a backend's function-level retries (Modal's
`retries=`) cover exceptions raised **inside** the container. They cannot
cover a spawn that failed before any container existed, an execution the
backend killed (OOM, timeout), a preempted worker, or one that died after
writing partial output. There was no `max_attempts` anywhere in the SDK, so
under `FAIL_FAST` one transient failure killed the whole build.

A tick is short-lived and remembers nothing, so the count comes from the
server: `FrontierTaskRef.attempt_count`, riding the frontier the tick already
fetches — no new round-trip on the happy path. Its window is the build's
current **round** (starts since the most recent `BUILD_RESUMED`), which is what
makes the recovery path the one operators already reach for.

## The default: `max_attempts = 2`

`1` would be the conservative choice if the failures on this path were a
random sample of failures — you do not want a deterministic bug run three
times. They are not. **An exception inside a task never reaches a tick at
all**: the worker self-reports `TASK_FAILED`, which takes the task out of the
frontier. The deterministic failures are therefore structurally out of reach
of this budget rather than excluded by a judgement call, which removes the
usual reason to default a retry policy to "off". What is left is the status
quo being replaced: under `FAIL_FAST`, one transient spawn failure kills an
entire build and the only recovery is a human noticing.

Not more than 2, because one retry covers the shape these failures actually
have (a spot preemption, an OOM on one worker, a backend that refused a spawn
for a moment); a task that fails the same way twice is telling you something
a third attempt will not fix; and 2 is the smallest step that closes the gap,
which is what a default that changes existing behaviour should be.

`max_attempts=1` restores the previous behaviour (record the failure, never
respawn).

## Which failures are retried

| Failure                                                             | Retried | Why                                                                                  |
| ------------------------------------------------------------------- | ------- | ------------------------------------------------------------------------------------ |
| `submit_detached` raised — no container ever ran                    | yes     | The backend has no function call to retry. This is the case nothing else covers.     |
| Backend reports the execution FAILED (OOM, kill, preemption)        | yes     | Transient by nature; the container is gone and its retry policy went with it.        |
| Execution claim lapsed with no ref to probe (worker vanished)       | yes     | Same failure seen from the claim side — see "claim expiry" below.                    |
| Task object missing and not rehydratable from registry data         | **no**  | Deterministic: the task store and the imported task classes do not change between two passes, so a retry re-reads the same absence and spends budget a genuinely transient failure elsewhere might have needed. |
| Exception inside your task                                          | n/a     | Never reaches a tick — the worker self-reports it. That is what `retries=` is for.   |

The verdict is passed explicitly at each failure site (`retryable=True/False`)
rather than derived centrally, so the split is legible at the point of
decision.

## Not fighting `FAIL_FAST`

`_handle_terminal` fails the build as soon as `status_counts["failed"] > 0`,
but those counts come from the frontier snapshot taken **before** the pass
acted. A failure recorded and then retried inside one pass is therefore never
counted as a build-killing failure: by the time the next frontier is fetched —
immediately, since the pass acted — the task is `PENDING` again. A failure
left failed (budget spent, or not retryable) appears in the very next snapshot
and fails the build promptly, exactly as before. No change to
`_handle_terminal` was needed.

The budget gate deliberately does **not** feed `denied_this_round`: that
number suppresses terminal detection because a limit-denied task is waiting
for a slot and will run. A budget-denied task will not run, and the failure
just recorded is what should fail the build.

## Rounds: what resets the budget, and what does not

A **round** runs from the build's most recent `BUILD_RESUMED` event, or from
the build's beginning. Two things that look alike from the UI are not:

- **Re-triggering the build resets it.** `_trigger_reactive` records
  `BUILD_RESUMED` *before* its discovery retries the failed tasks, so the
  round boundary reliably lands ahead of them and every task starts the new
  round at zero. This is the recovery path, and both exhaustion messages lead
  with it — with the build id filled in, and with
  `tick_kwargs={"max_attempts": N}` as the way to raise the budget on the same
  re-trigger.
- **A bare retry does not.** The UI's Retry, `stardag tasks retry` and the
  retry route flip the status without recording `BUILD_RESUMED`, so the count
  they are measured against is unchanged. On a task already at budget the retry
  *succeeds* and the scheduler still refuses to start it — which is exactly the
  trap the second message exists to name.

## Exhaustion is loud — both cases

**1. The tick declines to respawn (budget spent):**

> Task `<task-id>` of build `<build-id>` failed and will NOT be retried: its
> attempt budget for this build round is spent (2 of 2 allowed attempt(s)
> spent). To run it again, RE-TRIGGER THIS BUILD — build_trigger(...,
> build_id=`<build-id>`, reactive=True) — which starts a new round and resets
> every task's attempt count to zero. Retrying the task on its own does NOT
> reset the count: a bare retry (the UI's Retry, `stardag tasks retry`, the
> retry API route) makes the task pending again but leaves the budget spent, so
> this scheduler would decline to start it. If the task needs more attempts per
> round, re-trigger with a raised budget, e.g. tick_kwargs={"max_attempts": 4}.
> Last failure: `<reason>`

**2. A bare retry put a budget-exhausted task back in the frontier.** The
retry *succeeds* server-side and the task returns to `PENDING`, but it starts
no new round, so without this the operator would watch nothing happen:

> Task `<task-id>` of build `<build-id>` is PENDING again with its attempt
> budget for this build round already spent (2 of 2 allowed attempt(s) spent),
> so a BARE RETRY put it back — the UI's Retry button, `stardag tasks retry`,
> or the retry API route. That retry SUCCEEDED: the task really is pending
> again. But a bare retry does not start a new build round, so the attempt
> count it is measured against is unchanged, and the scheduler will not start
> it. Failing it again rather than leaving it PENDING and inert, so this is
> visible instead of looking like nothing happened. What you wanted is a
> RE-TRIGGER of this build — build_trigger(..., build_id=`<build-id>`,
> reactive=True) — which records BUILD_RESUMED, starts a new round, resets
> every task's attempt count to zero and re-runs exactly this task. Add
> tick_kwargs={"max_attempts": 4} to the re-trigger if it needs more attempts
> per round.

The task's recorded `error_message` carries the short form, so the explanation
is in the UI and not only in the tick's logs:

> Attempt budget spent (2 of 2 allowed attempt(s) in this build round). A bare
> retry does not reset it — re-trigger this build (build_id=`<build-id>`) to
> start a new round, optionally with tick_kwargs={"max_attempts": N}.

Both counts also ride the persisted `TickSummary`, flat and JSON-friendly like
the existing fields: `retried`, `retry_exhausted`, `budget_denied`.

## Interaction with claim expiry

Claim expiry (earlier in this stack) fails a task that is RUNNING with no ref
once its claim lapses — precisely the OOM/preemption case this PR exists for.
That path already returns `"failed"` from `_resolve_running`, so it arrives at
the same retryable branch as a probed-dead execution and is retried like any
other failure. The two mechanisms record **one** attempt between them, not
two: the expiry closes the attempt the frontier already counted, and the retry
records no start of its own — the next attempt is opened by the respawn.
Covered by a test asserting the count goes 1 → 2 across expire → fail → retry
→ respawn.

## Suspended resumption is never gated

A dynamic-dependency yield records a fresh start, so a task that yields
repeatedly accumulates attempts while being perfectly healthy. Gating
resumption would cap dynamic dependencies rather than retries, so the budget
gate applies to `PENDING` only. It does mean a suspend-heavy task reaches its
*retry* budget sooner within a round; the docstring says so and points at
raising `max_attempts`.

## Older servers

`attempt_count` is typed `int | None` (default `None`). A current server always
sends the field, so `None` means only one thing: a server predating it. That
distinction has to survive, because reading "no counter" as "0 attempts" makes
an unbounded retry loop unavoidable — with nothing counting, nothing can ever
stop allowing the next retry. So `None` disables retries and the tick logs why,
degrading to exactly the pre-`max_attempts` behaviour. `0` still never denies a
start.

## Also

- `max_attempts` joins the Modal `tick_kwargs` allowlist — the budget is per
  build, so a re-trigger is where a raised budget can be said.
- Docs: `concepts/build-execution.md` gains a "Task retries" section with a
  "Rounds: what resets the budget, and what does not" subsection;
  `how-to/integrate-modal.md` spells out that `FunctionSettings(retries=)` and
  `TickConfig.max_attempts` are different knobs for different failures and are
  meant to be set together, and shows the re-trigger that resets the budget.

## Verification

Rebased onto `perf/tick-fan-out` @ `3a6213c`.

`uv run pytest tests -q -m "not modal_live"` → **1214 passed** (1202 on the
base + 12 new), `uv run pyright src tests` → 0 errors, `pre-commit` clean on
every changed file.


andhus:
Copilot review triaged.

**`attempt_count == 0` comment.** Accepted and correct: the counter is scoped to the build's current *round* (reset on `BUILD_RESUMED`), so after a resume a task can read `0` having run in an earlier round of the same build. "Not attempted in this build" is misleading; it should say "not attempted in this round".

andhus:
**Now resolved** — the comment now says `attempt_count == 0` means "not attempted in this **round**", notes that BUILD_RESUMED resets it, and says why that is deliberate (a retry budget is per attempt at making the build progress; a resume is a new attempt).

PR #226, opened 2026-08-08
feat(api): make the SDK compatibility contract legible (and measurable)

andhus:
The API could not tell you which SDK was calling it, because the SDK never
said. This makes the SDK↔API compatibility contract **legible and
measurable** — and leaves enforcement switched off, because there is
currently nothing to enforce.

## This changes no client's behaviour

The API is wire compatible with every SDK released so far: the changes
since have been additive fields, parameters and endpoints, and no response
has changed shape or meaning. So **no minimum is set**, and with no minimum
set no request is ever rejected. What ships is the switch, off — plus the
measurement that makes flipping it later an informed decision rather than a
guess.

## The header (the bit the SDK PR keys on)

```
X-Stardag-SDK-Version: 0.18.0
```

Exact name: **`X-Stardag-SDK-Version`**. Value: the SDK's own version as a
PEP 440 string. Dedicated and machine-readable on purpose — a policy
decision must never depend on pattern-matching a `User-Agent`, which is
free-form and which proxies and instrumentation libraries rewrite without
telling anyone. The SDK is expected to also send a descriptive `User-Agent`
for logs; **the server never parses it** and nothing here depends on it.
`X-` prefixed to match the codebase's existing custom header (`X-API-Key`).

Treat the name as a wire constant: changing it breaks every released SDK's
ability to identify itself.

## The rejection contract (verbatim — this is what to write the SDK against)

When, and **only** when, a minimum is configured and the client explicitly
reports a version below it, the server responds:

- Status: **`426 Upgrade Required`**
- Body:

```json
{
  "detail": {
    "error_code": "SDK_VERSION_UNSUPPORTED",
    "message": "This Stardag server requires stardag SDK 0.18.0 or newer, but this request came from stardag 0.14.0. Upgrade with: pip install --upgrade \"stardag>=0.18.0\"  (the server's current requirement is always readable as minimum_sdk_version at GET /api/v1/version).",
    "sdk_version": "0.14.0",
    "minimum_sdk_version": "0.18.0"
  }
}
```

`error_code` is stable so a client can recognise "upgrade your SDK" without
parsing prose; the prose carries both version numbers and the command,
because the message is the user's entire experience of this feature.

Everything else is allowed:

| Client sends                    | Result                                        |
| ------------------------------- | --------------------------------------------- |
| nothing (every SDK in the wild) | **allowed**, logged                           |
| a malformed value               | **allowed**, logged — never a 400, never a 500 |
| `0.18.0.dev1` vs a `0.18.0` floor | **allowed**                                 |
| anything at all, no floor set   | **allowed** — the shipped default             |

A missing header being non-fatal is the load-bearing requirement: every SDK
released before the header existed sends nothing, so rejecting on absence
would break every existing client the moment this deploys.

## What's in it

- **`app/stardag-api/src/stardag_api/sdk_compat.py`** — the whole policy:
  header constant, parsing, comparison, the rejection body, and the
  first-seen recording. `SdkCompatSettings.minimum_version`
  (`STARDAG_API_SDK_MINIMUM_VERSION`) defaults to `None`. A misconfigured
  value fails at startup rather than silently enforcing nothing.
- **`middleware/sdk_version.py`** — applies to the whole surface without
  anyone remembering to opt a new route in (the SDK surface spans five
  routers, and a floor half the endpoints honour is worse than none), and
  sits outside the dependency graph so it is unaffected by auth. Sets
  `request.state.sdk_version` / `sdk_version_raw`. Registered before
  `CORSMiddleware` so CORS wraps it and a 426 still carries the headers a
  browser needs to read it.
- **Recording**: one log line per distinct version per process, capped
  against a client that randomises the value. The question worth answering
  is "which SDK versions are actually calling us?" — a line per request
  would bury it.
- **`GET /api/v1/version`** gains `minimum_sdk_version` (`null` by
  default). `/version` and `/health` are never gated, precisely so a client
  that was refused can find out what it is being asked to upgrade to.
- **Version comparison** is real PEP 440 via `packaging` — already
  installed transitively (gunicorn depends on it), now declared directly
  rather than relying on someone else's transitive dep. Pre-release, dev
  and local parts count as the base release: `0.18.0.dev1`, `0.18.0rc1` and
  `0.18.0+g1a2b3c` all satisfy a `0.18.0` floor. A developer running a
  local build of the very version we require must not be told to upgrade to
  it. The trade-off — an `rc` of X is trusted to behave like X — is
  documented.

## The process rule

Raising the floor breaks working deployments on purpose, so it carries an
obligation. **An API change that raises `minimum_sdk_version` must state it
in `CHANGELOG.md`, in `RELEASE_NOTES.md`, and in the error the server
returns.** That last one is automatic provided you set the value instead of
special-casing anything.

It lives in `DEV_README.md` under "Releasing the Server" → "Dropping
support for older SDKs", where someone cutting a release actually is, and
`CONTRIBUTING.md` points at it from the step where an API change gets
written — along with the expectation it follows from: keep Registry API
changes usable by SDKs already in the wild, prefer additive fields,
parameters and endpoints.

A newer SDK against an older self-hosted API is explicitly not a supported
configuration; self-hosters upgrade both together.

## Not included, deliberately

No "recommended version" soft band. Speculative while no floor exists at
all, and a second threshold to keep accurate is a real cost.

## Testing

`app/stardag-api/tests/test_sdk_version.py` — 46 tests. Absence and
malformed values first (the guarantee that would be expensive to get
wrong), then `/version` reporting `null` by default and the configured
value when set, `/version` and `/health` reachable by a refused client,
below/at/above the floor with a minimum configured in a fixture, dev and
pre-release builds of the floor, and a misconfigured floor failing at
construction.

```
$ uv run --no-sync pytest tests/ -q
514 passed, 29 skipped in 30.51s
```

`uvx pre-commit run --files <changed>`: prettier, ruff, ruff format and
pyright (app/stardag-api) all pass. The `app/stardag-ui` hooks (eslint,
tsc) can't run in this worktree — `node_modules` is not installed — so the
one-line type addition to `ServerVersion` was checked by inspection: the
interface is only used as the return type of `response.json()`, with no
object literals anywhere, so an added field cannot break the build.

---


andhus:
Copilot review triaged — three findings, all accepted, one behavioural.

1. **Trailing-slash variants of `/health` and `/api/v1/version`.** Real. The exemption is an exact path match, but FastAPI's `redirect_slashes` means a client can hit `/health/` — which the middleware gates with 426 *before* the redirect happens, breaking the stated guarantee that these endpoints are never blocked.
2. **`record_sdk_version()` returns early when the header is absent.** The important one: it contradicts the documented behaviour that a missing header is allowed but still observed, and it removes the ability to measure how many callers have not upgraded — which is the whole point of shipping the measurement before the enforcement.
3. **Double space in the rejection message.** Real, user-facing, and asserted on in tests.

andhus:
**Now resolved** — all three.

- **Absent header is recorded.** `record_sdk_version` returned early on `raw is None`, so the one cohort a floor most needs measuring — callers on an SDK too old to send the header — was invisible, defeating the point of shipping the measurement a release ahead of the enforcement. It gets its own `<absent>` label, deduped like any other so it cannot flood the log.
- **`/health/` and `/api/v1/version/` are exempt.** Note the live deployment cannot demonstrate this either way (no floor is configured there, so the middleware gates nothing) — it is covered by a unit test on the exemption set.
- Double space removed from the rejection message.

Verified: full API suite against **real Postgres** (558 passed), single Alembic head, and the merged stack redeployed to a live Modal+Neon environment.

PR #227, opened 2026-08-08
feat(sdk): identify the SDK version to the registry, and fail clearly when it is too old

andhus:
## Why

A registry can only tell an SDK **"you are too old"** if that SDK was already
identifying itself when it was released. A version that ships silent is
permanently un-diagnosable, and no later server change fixes that
retroactively. So this ships before there is anything to enforce — that is the
whole point of it.

The direction that matters is **old SDK against a new registry**: a hosted
registry always runs the latest API. The reverse (new SDK, old registry) is
explicitly *not* a supported combination — self-hosters upgrade both together —
so it gets a clear failure and no compatibility machinery. No
`MINIMUM_API_VERSION`, no eager `/version` handshake.

## What

**1. Every registry request identifies the SDK.** `X-Stardag-SDK-Version`
carries `stardag.__version__`, plus a conventional descriptive `User-Agent`
for logs only — the server keys on the dedicated header and deliberately
never parses the agent string. Both are set as *default headers on the httpx
clients* (sync, async, and the lock manager's), not at call sites, so a new
call site cannot forget them; httpx merges per-request `Content-Type` /
`Content-Encoding` over them. The version is resolved once at import.

`stardag.__version__` moved to a new leaf module `stardag._version` (and is
re-exported unchanged): the registry client needs it *while*
`stardag/__init__` is still executing, which rules out importing it from the
package root.

**2. `426 Upgrade Required` → `SDKVersionUnsupportedError`.** New in the
`stardag.exceptions` hierarchy (exported from `stardag`), carrying
`message`, `sdk_version` and `minimum_sdk_version`. The server's message
already names both versions and the exact `pip install --upgrade` line, so
it is surfaced **verbatim** rather than paraphrased — a paraphrase is a
second source of truth that drifts the day the server's wording changes.
The CLI prints it on its own line with rich markup off, not as a repr.

**3. Two confusing failures fixed.** `stardag builds cleanup` and
`stardag builds ticks` need endpoints an older registry lacks; FastAPI serves
an unknown path as the generic `{"detail": "Not Found"}`, which the CLI
rendered as "resource not found" — so version skew read as a bad build id.
Both now use `is_missing_route_error` (so a genuine resource-level 404 is
untouched) and say what is actually wrong:

```
Error: this registry does not support 'stardag builds ticks' — its
tick-summaries endpoint is missing. Upgrade stardag-api to a version
matching this SDK.

Error: this registry does not support 'stardag builds cleanup' — its
bulk-cancel endpoint is missing. Upgrade stardag-api to a version matching
this SDK.
Until then, cancel abandoned builds one at a time: stardag builds cancel
<build-id> --cascade
```

Codes against the server contract already implemented server-side.

## Tests

`1224 passed, 29 deselected` (base: 1214), `pyright src tests` clean,
pre-commit clean. New coverage: the header on sync and async requests and
surviving a per-request header merge; a 426 raising the typed error with the
server's message intact (structured and unstructured detail); both CLI
commands producing the upgrade message against a missing route, an app-level
404 still reading as not-found, and the 426 message printed as written.

`_CapturingRegistry` in the registry tests now swaps the *transport* rather
than replacing the whole client — otherwise the fixture would assert against
a client the SDK never builds.

## Docs

`SDKVersionUnsupportedError` in the exceptions reference, and a self-hosting
troubleshooting bullet covering skew in both directions.


andhus:
Copilot review triaged — both comments describe the same real gap.

**Non-JSON 426 loses its detail.** Correct. When `response.json()` fails (a proxy-generated 426 is the realistic case), `raw_detail` stays `None`, so `SDKVersionUnsupportedError` drops the only text the server or proxy actually sent and falls back to the generic message — in the one situation where the operator most needs the upstream's own words. Fix is to fall back to `response.text` in the parse-failure path.

andhus:
**Now resolved.**

Both 426 paths fall back to the text detail when the body is not JSON, so a proxy-generated 426 surfaces the upstream's own words instead of the generic upgrade sentence — which is what tells the reader *which hop* rejected them. Test added that genuinely discriminates (it fails without the fix).

PR #228, opened 2026-08-08
feat(reactive): discover the DAG inside Modal, not on the trigger's machine

andhus:
## The measurement

`_trigger_reactive` ran `discover_and_register_aio` **locally**. That walk
calls `task.complete_aio()` once per task — a target existence check. For a
`modalvol://` target root, from outside Modal, each of those is a Volume
**API** call, and they are rate limited.

Triggering a 127-task DAG from a laptop spent **~64 s**, almost entirely in
rate-limit backoff. Before that path was hardened it failed outright with
`VolumeListFiles rate limit exceeded`.

It was the only discovery site that ran locally:

| path | where discovery ran |
| --- | --- |
| `build_spawn` / `build_remote` (resident) | inside Modal — roots cloudpickled in |
| `_WorkerLifecycleReporter._register_dynamic_deps` | inside Modal (in the worker) |
| `_trigger_reactive` | **locally** |

Inside Modal the same volume is a **mounted filesystem**, so moving the walk
removes the API from the path rather than merely backing off against it.

## The new flow

```
trigger (cheap, no target I/O):
  mint or resume the build in the registry
  register the root tasks
  spawn bootstrap(build_id, roots-by-value, tick_kwargs)   -> return

bootstrap (in Modal):
  discover: walk requires(), one completion check per task
  register the discovered DAG (chunked, post-order)
  task-module coverage pre-flight
  persist the task objects (skipping pickles a tick can rebuild)
  write the reactive marker   <-- LAST
  spawn the first tick
```

The roots ride along **by value**, cloudpickled into the call exactly as
`build_spawn` passes `tasks=` to the builder.

Registering the roots is genuinely free of target I/O:
`_get_task_data_for_registration` reads `target().uri`, which *constructs* a
URI and performs no existence check. So the trigger touches no target root at
all — it now needs registry credentials only.

**The task-store writes move too, and that is a large part of the win.** With
a `modalvol://` root the store is a mounted filesystem in the container and a
rate-limited volume API from a laptop; the same N writes become N local ones.

## The ordering guarantee, and why it is preserved

`build_set_reactive_meta` is written **last**, after discovery *and*
persistence, and a tick no-ops on any build whose `reactive_app_name` is
None. That ordering is what guarantees **no tick ever observes a partially-
registered DAG**.

It is load-bearing, not stylistic. Registration is chunked (50 at a time)
post-order, so the roots land *last*; mid-registration a build presents as
"nothing actionable, roots not complete" — exactly the shape `_handle_terminal`
fails a build on. Today that is unreachable only because the marker is not yet
set.

The guarantee is enforced by keeping the marker write as the second-to-last
statement of `run_reactive_bootstrap` (the tick spawn follows it), and it is
called out in the function's docstring as *do not "tidy" this*, with the
failure mode spelled out — because it is easy to reorder and reintroduce the
window.

`BuildTriggerResult.function_call` now carries the **bootstrap** call for
reactive triggers. That is the honest handle: it is what the trigger actually
spawned, and its failure is what means the build never started. Previously it
carried the first tick's call, which resolved as soon as that tick lingered
out and said nothing about whether the DAG had been registered. The field name
and the type stay put, so resident callers are unaffected.

## Failure handling

A failed bootstrap must not leave an orphan RUNNING build, so both sides of
the spawn record a terminal `BUILD_FAILED` before propagating:

- **Trigger side** — from the moment it knows the build is RUNNING through the
  spawn. The arming point is deliberate: on a **re-trigger** the build may
  still be terminal until `build_resume` succeeds, so `build_resume` sits
  *outside* the guard. Failing a build on behalf of a resume that never landed
  would misattribute someone else's outcome.
- **Bootstrap side** — everything in the container, including the first tick's
  spawn. An un-spawned tick is not a partial success; it is a build nothing
  will ever move.

Recording the failure is best-effort and never masks the original exception.

## Where each check ended up

Discovery runs in Modal **by default, always**, with a named opt-out:
`StardagApp(reactive_discovery="modal" | "local")`. It sits on the app rather
than on the trigger because the reasons to opt out are properties of the
*deployment* — an app deployed before the `bootstrap` function existed, or a
target root the Modal app cannot reach — not of one invocation, and because
every other reactive knob already lives there. `"local"` runs the identical
`run_reactive_bootstrap` in the triggering process: same code, same order,
same failure handling.

The **authoritative** coverage pre-flight runs wherever discovery runs, over
the real discovered set, and gates `require_pickle_free`. It is never skipped.
A side benefit of the default placement: it now compares the DAG against the
**deployed** `task_modules` list rather than the caller's local one, closing
the stale-deploy blind spot the check used to carry.

`require_pickle_free=True` is still enforced and still fails loudly — it
records a terminal `BUILD_FAILED` and re-raises on the bootstrap's Modal call,
so it surfaces both in the registry and on `function_call.get()`.

Separately and additively, the trigger emits a **roots-only advisory** before
spawning: the root task classes are a fixed, tiny set it already holds, so the
check costs no `requires()` traversal, no target I/O and no measurable time,
and it catches the dominant misconfiguration ("I never declared my package")
in the operator's terminal rather than a container start later in a log they
have to go find. It is a strict *subset* of the real check and labelled as
such — two checks that can disagree would be worse than one. It is skipped
under `reactive_discovery="local"`, where the real check runs microseconds
later in the same process.

## Why a separate function rather than work inside the tick

The timeout budgets answer different questions. A tick is sized for **one
frontier pass** and is meant to be short — its timeout also derives the
per-pass spawn cap. The bootstrap is a **single whole-DAG walk** whose cost
scales with the DAG and is paid once per trigger. One number cannot honestly
cover both: shortening the tick, normally a good idea, would start killing the
bootstrap of large DAGs.

`bootstrap` therefore defaults to `builder_settings` (**not** `tick_settings`)
— the same image, secrets and target-root volume mounts as the builder, which
does the same discovery for resident builds — and is separately configurable
via `StardagApp(bootstrap_settings=...)`.

Resident builds are completely unaffected.

## Verification

```
uv run --project lib/stardag pytest lib/stardag/tests -q -m "not modal_live"
1247 passed, 29 deselected, 15 warnings in 24.80s        (1227 on the base)

uvx pre-commit run --files <every changed file>
prettier ... Passed   ruff ... Passed   ruff format ... Passed   pyright (lib/stardag) ... Passed
```

Deploying an app is required to get the `bootstrap` function.


andhus:
Copilot review triaged — both findings are real and both are about `reactive_discovery="local"`.

**`BuildTriggerResult.function_call` docstring.** Correct: with `reactive_discovery="local"` the trigger does not spawn `bootstrap` and returns the first tick's `FunctionCall`.

**`_build_executor_metadata` records `function_name="tick"`.** The sharper one — the trigger now spawns `bootstrap` by default, so the stored metadata names a function that did not run, and that field is read by operator/UI surfaces.

andhus:
**Now resolved** — both.

`_build_executor_metadata` records `bootstrap` for reactive builds, `tick` under `reactive_discovery="local"` (no bootstrap to spawn), `build` otherwise. The `function_call` docstring gains the same exception.

Confirmed on a live deployment: a reactive build triggered after this change records `function_name=bootstrap`.

